PR #3, opened 2018-07-22
Add appveyor support

pfmoore:


pfmoore:
There's a failure in Python 2.7 - see https://ci.appveyor.com/project/pfmoore/pep517/build/job/gq9h9qabohh03kt2

It looks like it might be something in the `testpath` library at first glance (the `modified_env`) decorator) although I can't immediately see what would be wrong there.

pfmoore:
By the way - it looks like the travis config is relying on some preinstalled packages (I needed to add pytest and mock to get the tests to work). It might be worth adding an explicit `dev-requirements.txt` file, and/or instructions on how to run the tests. I'll do that in a follow-up PR once I get the Appveyor tests working.

pfmoore:
I'm looking into it now. I wonder why it doesn't show up on Unix?

takluyver:
Maybe it's retrieving a path from somewhere which returns unicode on Windows and bytes on Linux, corresponding to the platforms' native path types?

pfmoore:
OK, `pytoml.load` returns Unicode strings on Python 2.7. On Windows under Python 2.7, it looks like the `env` argument for `subprocess.call` requires `str` values (it calls CreateProcess under the hood, which is where the difference with Unix arises). I've just pushed what's hopefully a fix (that has more comment than actual code change - encoding issues are always messy to explain!)

pep517/wrappers.py
```diff
@@ -110,7 +110,23 @@ def build_sdist(self, sdist_directory, config_settings):
 
     def _call_hook(self, hook_name, kwargs):
         env = os.environ.copy()
-        env['PEP517_BUILD_BACKEND'] = self.build_backend
+
+        # On Python 2, pytoml returns Unicode values (which is correct)
+        # but the environment passed to check_call needs to contain string
+        # values. We convert here by doing a default encode() which will
+        # use the system default encoding. That should do the right thing,
```
takluyver: `unicode.encode()` actually uses Python's default encoding, IIRC, not the system default. So on Python 2, it's equivalent to `.encode('ascii')`, unless someone has done the tricks necessary to call `sys.setdefaultencoding()` at startup, which is not recommended.

takluyver:
I think the actual code change is OK, it's only the comment I'm quibbling with :-)

pfmoore:
Thanks for the review. I'll actually change the code to use `sys.getdefaultencoding()` because while ASCII should be fine, the PEP *technically* doesn't specify encoding, it just says "letter" and I wouldn't want a horde of non-English speakers yelling at me about their letters ;-)

takluyver:
`sys.getdefaultencoding()` is also Python's default encoding, i.e. always ASCII for Python 2. If you want the locale encoding from the platform, I think that's `locale.getpreferredencoding()` (which, incidentally, is not threadsafe on all platforms).

However, since the `build_backend` refers to Python modules and objects, it has to be ASCII for Python 2 anyway, because Python 2 identifiers can't have non-ascii characters. So I'd just make it explicitly `'ascii'`.

pfmoore:
Bah, you're right. I'd forgotten how annoying Python 2's Unicode handling is...

takluyver:
:+1:

pfmoore:
Thanks! I always like it when adding CI actually does pick up an issue to solve - makes it feel worthwhile :-)

PR #7, opened 2018-07-25
Change Pep517HookCaller to not read pyproject.toml directly

pfmoore:
Modify the low-level wrappers to leave reading of `pyproject.toml` to the caller.

When trying to use the wrappers from pip, the fact that the wrappers do their own parsing and validation of `pyproject.toml` clashes with pip's own parsing and validation. Rather than trying to get the two to work together, it seems to me better to take the approach that it's the caller's responsibility to parse the `pyproject.toml` file, and pass the relevant information (the backend name and the source directory) to the wrappers, rather than have the wrappers try to do two unrelated jobs.

I'm not going to merge this change immediately, as it's a change to the external API, and I'd like some feedback on whether it's OK.

### DIFF
--- a/pep517/check.py
+++ b/pep517/check.py
@@ -4,7 +4,7 @@ import argparse
 import logging
 import os
 from os.path import isfile, join as pjoin
-from pytoml import TomlError
+from pytoml import TomlError, load as toml_load
 import shutil
 from subprocess import CalledProcessError
 import sys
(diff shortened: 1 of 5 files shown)


takluyver:
I'm not aware of any external consumers of this project yet, so I think whatever makes sense for pip is OK.

It does have an internal consumer in the `pep517.check` module which may need a minor change.

pradyunsg:
> It does have an internal consumer in the pep517.check module which may need a minor change.

It's been updated.

takluyver:
Oh, thanks, I missed that.

PR #21, opened 2018-12-11
Delint

jaraco:
This commit simply reformats the main codebase to conform to linter conventions to avoid warnings in my IDE. If you'd rather I suppress the various warnings, let me know and I'll add a .flake8 with the indicated rules.


jaraco:
> I think we should also add flake8 to CI for this project.

Thanks. I'll consider that for a subsequent contrib.

PR #27, opened 2018-12-12
Add build command

jaraco:
This PR introduces `pep517.build`, modeled after `pep517.check`. Fixes #25.

As @pfmoore mentioned in that ticket, the README indicates the experimental nature of this first implementation of a build tool that can build an sdist for PEP 517 projects, and is the first that makes it possible to release some projects without invoking easy_install.

PR #32, opened 2018-12-17
Automatically cut releases from tagged commits

jaraco:
This commit adds a new 'job' to the Travis config, one which releases tagged commits, based on work I've done in jaraco/skeleton and which I use in hundreds of other projects to streamline the release process.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+dist: xenial
+sudo: false
 language: python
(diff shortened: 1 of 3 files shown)


jaraco:
This automated release process isn't working because #27 isn't released yet... and I wanted to cut an 0.4 before merging that functionality. :(

PR #46, opened 2019-03-16
Support in tree hooks (pyproject.toml backend-path key)

pfmoore:


pep517/_in_process.py
```diff
@@ -29,6 +30,23 @@ def __init__(self, traceback):
         self.traceback = traceback
 
 
+class BackendInvalid(Exception):
+    """Raised if we cannot import the backend"""
```
takluyver: Copy-pasted docstring?

pep517/_in_process.py
```diff
@@ -29,6 +30,23 @@ def __init__(self, traceback):
...
+    def __init__(self, message):
+        self.message = message
+
+
+def contained_in(filename, directory):
+    """Test if a file is located within the given directory."""
+    filename = os.path.normcase(os.path.abspath(filename))
+    directory = os.path.normcase(os.path.abspath(directory))
+    return os.path.commonprefix([filename, directory]) == directory
```
takluyver: `commonprefix()` appears to be a purely string-based operation, with no awareness of filesystems. So is `commprefix([a, b]) == b` any different from `a.startswith(b)`?

pep517/_in_process.py
```diff
@@ -197,12 +223,23 @@ def main():
 
     hook_input = compat.read_json(pjoin(control_dir, 'input.json'))
 
+    # Add in-tree backend directories to the front of sys.path.
+    # The variable extra_path is global so that it can be used within
+    #  _build_backend.
+    global extra_path
+    extra_path = hook_input.get('backend_path')
```
takluyver: I think it would be neater to do this as an environment variable from the frontend, like we already do `PEP517_BUILD_BACKEND`, so we avoid the main() function having to put it in a global variable. That keeps all the import-related code together in `_build_backend()`.

pep517/wrappers.py
```diff
@@ -41,15 +49,45 @@ def default_subprocess_runner(cmd, cwd=None, extra_environ=None):
...
+        raise ValueError("paths must be relative")
+
+    abs_source = os.path.abspath(source_tree)
+    abs_requested = os.path.abspath(os.path.join(abs_source, requested))
+    # We have to use commonprefix for Python 2.7 compatibility. So we
+    # normalise case to avoid problems because commonprefix is a character
+    # based comparison :-(
+    norm_source = os.path.normcase(abs_source)
+    norm_requested = os.path.normcase(abs_requested)
+    if os.path.commonprefix([norm_source, norm_requested]) != norm_source:
```
takluyver: Ditto re commonprefix

pep517/_in_process.py
```diff
@@ -37,6 +55,14 @@ def _build_backend():
         obj = import_module(mod_path)
     except ImportError:
         raise BackendUnavailable(traceback.format_exc())
+
+    if extra_path:
+        for path in extra_path:
+            if contained_in(obj.__file__, path):
+                break
+        else:
+            raise BackendInvalid("Backend was not loaded from backend-path")
```
takluyver: I think this check could be more concise with `if not any(...)`

pep517/wrappers.py
```diff
@@ -41,15 +49,45 @@ def default_subprocess_runner(cmd, cwd=None, extra_environ=None):
...
+    Ensure that the requested backend path is specified as a relative path,
+    and resolves to a location under the given source tree.
+
+    Return an absolute version of the requested path.
+    """
+    if os.path.isabs(requested):
+        raise ValueError("paths must be relative")
+
+    abs_source = os.path.abspath(source_tree)
+    abs_requested = os.path.abspath(os.path.join(abs_source, requested))
```
takluyver: After joining to `abs_source`, this should already be absolute in all cases, shouldn't it? Maybe `normpath` makes it clearer what is happening.

tests/samples/pkg_intree/backend/intree_backend.py
```diff
@@ -0,0 +1,2 @@
+def get_requires_for_build_sdist(config_settings):
+    return []
```
takluyver: Maybe return some dummy value just so the tests can check it's really this function being called.

pep517/_in_process.py
pfmoore: Yep, will fix.

pep517/_in_process.py
pfmoore: Correct, it's a string-based operation. If we didn't need compatibility with older Python versions, I'd use commonpath (or pathlib). If you know of a Python 2 compatible way of saying "is this filename in this directory", I'd happily use it. But in the absence of anything better, I prefer `os.path.commonprefix` because at least it more clearly indicates the intent.

pep517/_in_process.py
pfmoore: I'll give that a try. The only problem I have is that it's a list, so there's a translation step between list and string - I guess I can use `os.pathsep` for that, though.

pep517/_in_process.py
pfmoore: Good idea - will fix.

pep517/wrappers.py
pfmoore: True. Will fix.

tests/samples/pkg_intree/backend/intree_backend.py
pfmoore: Done

pep517/_in_process.py
pfmoore: I've made this change, but in my opinion it's a lot less clean. The problem is that we need to consider all of the following possibilities:

* backend-path might not exist
* backend-path is a list, so we need to join it into a string
* encodings - path encodings are difficult, especially on Unix...

Overall, I think my original approach is better, but let me know what you think (and if you have any suggestions for making either version cleaner). Unless you're really keen on the environment variable version, I think I'll revert that and go back to the global variable.

Another option would be to make the backend a class, and the hooks methods on that class, and stop using globals altogether. But that's a bigger restructuring, and maybe should be left for a separate PR?

takluyver: I'm still inclined to prefer the version with the environment variable, but if you're really keen to go back, I'll stop arguing about it.

The only point that really presents difficulties is the encoding question, and I don't think putting it in JSON makes that any easier. Paths on Unix are fundamentally bytes, whereas JSON strings are fundamentally unicode (or UTF-16, to be more precise). Since environment variables are also bytes, we could do this neatly for Unix using `os.environb` and bytes paths. But of course that wouldn't be straightforward for Windows.

pfmoore: I'm willing to leave it in (it's honestly just easier than taking it out again). But I'll note the following point for the record, it's not just (or even mostly) the encoding that's the issue here for me. It's that JSON does the job of serialising Python data structures, so that we don't have to. With JSON, we just pass the `backend_path` list and know it'll arrive at the other side unscathed. By using an environment variable, we have to do all of that serialisation ourselves.

takluyver: 50% devil's advocate: the two aren't mutually exclusive. As we control both sides, we could put a JSON list in an environment variable. :wink:

pfmoore: Sigh. I wondered if you'd suggest that :-P Still leaves the encoding issue, though.

I'm OK with leaving it as it is now, assuming you are. Are there any other points you think I need to address, or is this ready?

takluyver: I'd modify `sys.path` inside `_build_backend()`, rather than in `main()`. That was the main reason I preferred the environment variable: it means all the logic for finding and loading the backend can live together in `_build_backend()`.

pfmoore: Ah, I see what you were meaning now - yes, that does make the whole code path a lot simpler. (I was distracted by a general principle that I heard somewhere that you should only modify `sys.path` at the start of your program - but even if that is a good principle, it makes no difference in this case).

I'll update the code on that basis, it might not be until next weekend, though.

pfmoore:
@takluyver I believe that's all the review comments addressed. Sorry it took so long. Are there any other issues you think need addressing?

pep517/wrappers.py
```diff
@@ -143,25 +181,40 @@ def _call_hook(self, hook_name, kwargs):
...
         if sys.version_info[0] == 2:
             build_backend = self.build_backend.encode('ASCII')
         else:
             build_backend = self.build_backend
+        extra_environ = {'PEP517_BUILD_BACKEND': build_backend}
+
+        if self.backend_path:
+            backend_path = os.pathsep.join(self.backend_path)
+            if sys.version_info[0] == 2:
+                backend_path = backend_path.encode(sys.getfilesystemencoding())
```
takluyver: This may fail if a Unix filesystem erroneously reports that it's ASCII and someone tries to use a non-ASCII path. But I don't know what the correct thing to do on Python 2 in that case is (Python 3 has `os.fsencode()`). So we may as well cross our fingers and hope people use either Python 3 or plain ASCII paths.

takluyver:
Thanks, I think it looks good.

PR #47, opened 2019-05-18
Tests are run with tox

jaraco:
I suggest the readme should refer to the more superior mechanism to run tests and link to the project for more info.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -11,7 +11,7 @@ provides:
 - Higher-level functions which install the build dependencies into a
   temporary environment and build a wheel/sdist using them.
 
-Run the tests with `tox <https://pypi.org/project/tox>`_.
+Run the tests with ``pytest`` or `tox <https://pypi.org/project/tox>`_.
 
 High level usage, with build requirements handled:
 
--- a/README.rst
+++ b/README.rst
@@ -11,7 +11,7 @@ provides:
 - Higher-level functions which install the build dependencies into a
   temporary environment and build a wheel/sdist using them.
 
-Run the tests with ``py.test``.
+Run the tests with `tox <https://pypi.org/project/tox>`_.
 
 High level usage, with build requirements handled:
 


takluyver:
I'm somewhat conflicted about this. I accept that tox is 'superior' in that it covers more possibilities. But pytest is what I almost always actually run while developing stuff, because it's typically an order of magnitude faster and catches >=95% of the problems that tox would have done.

I guess tox is the right thing to point to for people who aren't already familiar with a project, because it handles more of the setup. But on the other hand, it's another layer of abstraction between the programmer and the code: the more tools you need to work on some code, the more offputting it is for people not already familiar with those tools.

pfmoore:
I agree, and I think it's nice that this project is set up so that you *can* run the tests with a simple `py.test` (not all projects are).

Why not just include both ("run the tests with `py.test` (or `tox` if you prefer)")?

takluyver:
That's probably reasonable. I'm also sometimes picky about giving too many options in introductory documentation, because options demand effort from the reader. But I quite like your wording: there's a clear default for people who don't know which they prefer.

jaraco:
> the more tools you need to work on some code, the more offputting it is for people not already familiar with those tools.

I agree fervently, which is why I recommend `tox`. One doesn't need both `pytest` (and which version) and `tox`, as `tox` will facilitate the creation of pytest. Tox will even guarantee the required version is installed (if it wasn't invoked under the required version). It's also test-runner agnostic, meaning it can run nose or unittest or pytest or custom tests. That's why I recommend that those developing on Python have tox available. If you don't have it available for this project, you'll probably need it for the next.

For me, it's a clear win. Let the project control the test runner (including required version) and let the user simply run `tox`.

For optimized workflows, like the one you describe, where you have already addressed all of the issues that `tox` adjusts, you can simply ignore the advice in the readme. That advice is for beginners, not for more optimized workflows.

> Why not just include both ("run the tests with py.test (or tox if you prefer)")?

I don't like that approach as it's ambiguous, but also more prone to error. My preference would be to advertise the [preferably one obvious way](https://www.python.org/dev/peps/pep-0020/) and that way should be the same technique that's employed in CI.

---

I don't feel strongly about it. If you're not convinced, please feel free to close.

jaraco:
This change also fixes a bug that `py.test` is [deprecated](https://stackoverflow.com/a/41893170/70170) (in favor of `pytest`) and adds a hyperlink to the tool to help someone unfamiliar to get started.

takluyver:
Amusingly, you changed it just as I was typing a message to say I'd accept it as it was. But I'm happy with it this way.

Issue #26, opened 2018-12-12
Expose command to build package metadata


jaraco:
In pypa/packaging-problems#224 and #19, @JonasT and I are trying to work out how to use PEP 517 to query the package metadata for an unbuilt package. PEP 517 has an interface for `prepare_metadata_for_build_wheel` and a protocol for falling back to `build_wheel` to generate that metadata.

This package should present a command that will follow that protocol for producing such metadata from a source checkout, similar to what `pep517.check` does. It should probably present both as a command, leaving the metadata somewhere on the filesystem, and as an API - returning a path to the generated metadata.

jaraco:
In pypa/packaging-problems#226, I propose an extension to the PEP 517 API that would enable tools like this package to build the package metadata and return a data stream or object in memory without having to rely on a file system.

Issue #13, opened 2018-08-17
Callers need better control over backend output


pfmoore:
From the PEP, "The build frontend may capture stdout and/or stderr from the backend". As a specific example, pip currently captures setuptools output and displays it only in debug mode or when an error occurs. Also, if not displaying output, a progress spinner is updated as lines of output are produced. Similar control will be needed under PEP 517.

So, the requirements are, to be able to (on a per-call basis) capture hook output (stdout and/or stderr - pip simply merges the two, which should be sufficient for most cases) and react to it in real time as it's generated line by line.

It's not immediately obvious what the best design here would be. My initial suggestion (not yet validated to confirm that it would satisfy pip's requirements) would be for the wrapper to support a context manager that allows temporarily using a different function to call the subprocess:

```python

def my_runner(cmdline, cwd, env):
    # This is the default, we'd customise this as needed
    # Callers must supply a function that runs the command with the given CWD and env.
    # No return value.
    # Raise an exception if the subprocess errors, or the invocation fails.
    subprocess.check_call(cmdline, cwd, env)

with backend.subprocess_runner(my_runner):
    backend.build_wheel(...)
```

Does anyone have any comments, or alternative design ssuggestions?

pfmoore:
Thanks. Having tested it in pip, I think the right granularity is being able to replace `check_call(cmd, cwd, extra_env)` is about right (although that may just be bias because it matches pip's use case :smile:). But I'm not a huge fan of the context manager, it feels a bit over-engineered. A `runner` kwarg is not unreasonable, but I wasn't sure if I liked changing the calling signatures of the hooks.

See https://github.com/pfmoore/pip/blob/pep517/src/pip/_internal/wheel.py#L682 for how it looks with the context manager version. (Warning: that branch is actively having history rewritten at the moment, so that URL may not be stable long term).

takluyver:
For now, I think whatever makes sense for pip is a good way to go. If we need more flexibility in the future, I like my last idea (a lower level API that prepares argv+env+cwd but doesn't run the subprocess).

pfmoore:
OK, cool. I'm not entirely sure I like the lower-level API (the wrappers are currently really nice, insofar as they feel like you're just calling an in-process API, and I think that's the right level for anything but advanced use) but it depends on what level of control the caller wants. Pip wouldn't want to go any lower level than the current API, as if we did, there'd be little point in not just making our own wrappers, but maybe other projects would want something different.

[...]

PR #48, opened 2019-05-21
New command pep517.meta

jaraco:
As outlined in #26, this command implements a ~`pep517.build-meta`~`pep517.meta` command, allowing for building metadata for a source directory.

The new ~`build_meta`~`meta` module also exposes an API for retrieving the metadata as an in-memory zip file allowing consumers to receive metadata in-process without thread safety and other race concerns.

Additionally, this request adds a `compat_system` function which, given a source directory, produces a "build system", either by loading it from pyproject.toml or falling back to "setuptools (legacy)".

This proposed technique fulfills the needs of `importlib_metadata` or other consumers that wish to load metadata from PEP 517 projects on disk, or if `compat_system()` is passed for `system`, for most any project.

Edit: This technique additionally introduces a `load()` function for loading an `importlib_metadata.Distribution` object from a source directory.

One downside I'd like to still investigate:

- [x] When generating the metadata, a lot of output is emitted. It would be nice to be able to suppress that output.


jaraco:
The code in b35a8dd is currently released to https://m.devpi.net/jaraco/pep517-meta.

jaraco:
See [importlib_metadata!67](https://gitlab.com/python-devs/importlib_metadata/merge_requests/67) for the corresponding importlib_metadata work.

pfmoore:
If I understand correctly, this implements a proposed extension to PEP 517? If that's true, then I think that we should hold off on merging this PR until that proposed extension is accepted,

jaraco:
> this implements a proposed extension to PEP 517?

No. This PR only implements APIs around the already specified hooks defined in PEP 517, namely [prepare_metadata_for_build_wheel](https://www.python.org/dev/peps/pep-0517/#prepare-metadata-for-build-wheel).

This new command and API follow the patterns found in `pep517.build` and `pep517.check` to present comparable functionality for metadata.

It does go one step further. Based on the discussion and recommendation in pypa/packaging-problems#226, the `build_meta_as_zip` function builds the metadata in a temporary directory, zips it up in-memory, and then returns that in-memory zipfile.

But as you can see, each of these functions are disentangled from the more fundamental functionality (building the metadata in some target directory).

pfmoore:
>> this implements a proposed extension to PEP 517?
>
> No.

OK, cool. My apologies, I got the idea it was an extension from [this comment](https://github.com/pypa/pep517/issues/26#issuecomment-446723506). I must have misread the thread (I was skimming quickly, I don't have time right now for a full review).

pep517/build-meta.py
```diff
@@ -0,0 +1,3 @@
+from . import build_meta
+
+__name__ == '__main__' and build_meta.main()
```
gaborbernat: wow I've never seen this cryptic way of doing this 🤔

jaraco:
Here's an example of this functionality in action:

```
~ $ pip-run -i https://m.devpi.net/jaraco/pep517-meta pep517                                                                                                                   
Looking in indexes: https://m.devpi.net/jaraco/pep517-meta
Collecting pep517
  Downloading https://m.devpi.net/jaraco/pep517-meta/%2Bf/1f7/c95a2a4d63948/pep517-0.5.0-py2.py3-none-any.whl (49kB)
     |████████████████████████████████| 51kB 440kB/s 
Collecting pytoml (from pep517)
Installing collected packages: pytoml, pep517
Successfully installed pep517-0.5.0 pytoml-0.1.20
>>> Python 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 16:52:21) 
[Clang 6.0 (clang-600.0.57)] on darwin
Type "help", "copyright", "credits" or "license" for more information.

>>> import pep517.build_meta
>>> import functools
>>> import os
>>> builder = functools.partial(pep517.build_meta.build_meta, source_dir=os.path.expanduser('~/m/jaraco.functools'))
>>> meta_zip = pep517.build_meta.build_meta_as_zip(builder)
Collecting setuptools>=34.4
  Using cached https://files.pythonhosted.org/packages/ec/51/f45cea425fd5cb0b0380f5b0f048ebc1da5b417e48d304838c02d6288a1e/setuptools-41.0.1-py2.py3-none-any.whl
Collecting wheel
  Using cached https://files.pythonhosted.org/packages/bb/10/44230dd6bf3563b8f227dbf344c908d412ad2ff48066476672f3a72e174e/wheel-0.33.4-py2.py3-none-any.whl
Collecting setuptools_scm>=1.15
  Using cached https://files.pythonhosted.org/packages/1d/70/97966deebaeeda0b81d3cd63ba9f8ec929b838871ed17476de9d8159db3e/setuptools_scm-3.3.3-py2.py3-none-any.whl
Installing collected packages: setuptools, wheel, setuptools-scm
Successfully installed setuptools setuptools-scm wheel
Collecting wheel
  Using cached https://files.pythonhosted.org/packages/bb/10/44230dd6bf3563b8f227dbf344c908d412ad2ff48066476672f3a72e174e/wheel-0.33.4-py2.py3-none-any.whl
Collecting setuptools_scm>=1.15.0
  Using cached https://files.pythonhosted.org/packages/1d/70/97966deebaeeda0b81d3cd63ba9f8ec929b838871ed17476de9d8159db3e/setuptools_scm-3.3.3-py2.py3-none-any.whl
Installing collected packages: wheel, setuptools-scm
Successfully installed setuptools-scm-3.3.3 wheel-0.33.4
running dist_info
creating /var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T/tmp04lgx4m_/jaraco.functools.egg-info
writing /var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T/tmp04lgx4m_/jaraco.functools.egg-info/PKG-INFO
writing dependency_links to /var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T/tmp04lgx4m_/jaraco.functools.egg-info/dependency_links.txt
writing requirements to /var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T/tmp04lgx4m_/jaraco.functools.egg-info/requires.txt
writing top-level names to /var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T/tmp04lgx4m_/jaraco.functools.egg-info/top_level.txt
writing manifest file '/var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T/tmp04lgx4m_/jaraco.functools.egg-info/SOURCES.txt'
writing manifest file '/var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T/tmp04lgx4m_/jaraco.functools.egg-info/SOURCES.txt'
creating '/var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T/tmp04lgx4m_/jaraco.functools.dist-info'
>>> meta_zip
<zipfile.ZipFile file=<_io.BytesIO object at 0x10f9985c8> mode='w'>
```

Another thing I'm considering is instead of stopping at generating the zip file, perhaps `pep517` could own [this functionality](https://gitlab.com/python-devs/importlib_metadata/blob/35189f591fa736da4a28cdb20b34cf9be02c2918/importlib_metadata/__init__.py#L200-206). Then, `importlib_metadata` would facilitate the metadata parsing, but `pep517` could own the metadata loading logic. That feels like a better separation of concerns, especially as `importlib_metadata` enters the stdlib and can't reliably depend on `pep517`.

pep517/build_meta.py
```diff
@@ -0,0 +1,82 @@
...
+            rel_path = os.path.relpath(fs_path, root)
+            zip_file.writestr(rel_path + '/', '')
+        for path in files:
+            fs_path = os.path.join(root, path)
+            rel_path = os.path.relpath(fs_path, root)
+            zip_file.write(fs_path, rel_path)
+    return zip_file
+
+
+def build_meta_as_zip(builder=build_meta):
```
gaborbernat: this is nowhere used, what's the goal with this?

jaraco:
Also, as is demonstrated by failures I discovered above, we should add tests.

- [ ] Add tests for this functionality.

pep517/build_meta.py
```diff
@@ -0,0 +1,82 @@
...
+    with tempdir() as out_dir:
+        builder(dest=out_dir)
+        return dir_to_zipfile(out_dir)
+
+
+parser = argparse.ArgumentParser()
```
gaborbernat: would prefer this to be not global (aka let's put it inside a function, and call it at the start of main); will make it easier to test

pep517/dirtools.py
```diff
@@ -0,0 +1,25 @@
+import os
+import contextlib
+import tempfile
+import shutil
+import errno
+
+
+@contextlib.contextmanager
+def tempdir():
```
gaborbernat: this deserves some documentation

pyproject.toml
```diff
@@ -10,6 +10,8 @@ home-page = "https://github.com/takluyver/pep517"
 description-file = "README.rst"
 requires = [
     "pytoml",
+    "importlib_metadata",
```
gaborbernat: why is this a build dependency? 🤔

jaraco: I thought that was an install dependency.

pfmoore: This one's the flit metadata section, not the build-system section.

It's a bit unfortunate the name's used in 2 places for 2 different things :-(

pep517/build_meta.py
jaraco: I agree, except this code follows the pattern of the `build` and `check` modules. It's more important for these to be consistent. If we wish to change the pattern (a reasonable request), I suggest doing that subsequently to accepting this addition.

pep517/build_meta.py
jaraco: The latest commits use this code. It's meant to supply the metadata as the return value of the function. Because the metadata is inherently a multi-file structure, it has to be a zip file.

pep517/build-meta.py
jaraco: I agree it's idiosyncratic, but I've run it by other seasoned devs and gotten at least a nod of endorsement about the one-line approach, so I've started using it. I personally prefer it to the traditional approach for two reasons: (1) it fits on one line, and (2) it encourages exactly one expression for `__main__`. What I'd really prefer is a builtin that does what [autocommand](https://pypi.org/project/autocommand) does - decorate a function to be run as main. I use that wherever I can rely on Python 3 and third-party dependencies.

pfmoore: I'm -1 on this. Let's keep to the standard `if __name__ == '__main__'` idiom. Why make it harder for future maintainers to understand for little benefit?

gaborbernat: I definitely find it hard to read, but that's only me perhaps.

pep517/build_meta.py
gaborbernat: You're correct.

pep517/dirtools.py
jaraco: Done.

pep517/build-meta.py
jaraco: I updated this and ultimately removed the file entirely after I found that `pep517.meta` was a more suitable name for both the command and the API module.

jaraco:
As you can see, tests are failing on Python 2.7 due to `flit` not being available on Python 2.7 (and as a result, it's not possible to build the metadata for `pep517` on Python 2.7). I guess that's an xfail?

jaraco:
In e3138f47e3a921abe7627087c1a57ab82023c79b and d27a2eceb41f403e437b06e314e2e81ed6dbe7f0, I added a test and updated the implementation to ensure that the pip install and build steps don't emit any output. The latter change has implications for other parts of the library that also call `pip_install` (their output will be captured/suppressed also). For this reason alone, I'm uneasy about this change.

I'd like feedback from others monitoring this thread on those concerns.

jaraco:
I feel a lot better about this latest implementation. It suppresses the output from the `pip_install` invocation by redirecting the output to the log.

pep517/build.py
```diff
@@ -3,25 +3,50 @@
...
 
 
-@contextlib.contextmanager
-def tempdir():
-    td = tempfile.mkdtemp()
+def validate_system(system):
+    """
+    Ensure build system has the requisite fields.
+    """
+    required = {'requires', 'build-backend'}
```
gaborbernat: seems overzealous, can't we fallback for the build-backend key to setuptools legacy as pip does?

gaborbernat: I would say even for requires we should just inject ``setuptools >= 41.0.0`` and ``wheel >= 0.30.0``.

pep517/build.py
```diff
@@ -3,25 +3,50 @@
...
+        missing = required - set(system)
+        message = "Missing required fields: {missing}".format(**locals())
+        raise ValueError(message)
+
+
+def load_system(source_dir):
+    """
+    Load the build system from a source dir (pyproject.toml).
+    """
+    pyproject = os.path.join(source_dir, 'pyproject.toml')
```
gaborbernat: again let's fallback to old things rather than fail if missing 👍

pep517/envbuild.py
```diff
@@ -90,9 +90,14 @@ def pip_install(self, reqs):
         if not reqs:
             return
         log.info('Calling pip to install %s', reqs)
-        check_call([
+        cmd = [
             sys.executable, '-m', 'pip', 'install', '--ignore-installed',
-            '--prefix', self.path] + list(reqs))
+            '--prefix', self.path] + list(reqs)
+        check_call(
```
gaborbernat: will this print out the output if the call fails? 🤔

tests/test_meta.py
```diff
@@ -0,0 +1,38 @@
+from __future__ import unicode_literals, absolute_import, division
+
+import re
+
+import pytest
+
+from pep517 import meta
+
+
+pep517_needs_python_3 = pytest.mark.xfail(
```
gaborbernat: what 🤔 why does this requires python 3? 🤔

pep517/build.py
jaraco: This function is simply about ensuring the requisite fields are there. There is another function, `compat_system`, that supplies the fallback. If we could say that "there is no case where the fallback would not be desirable", then I agree, the fallback should be derived unconditionally and there's no need to validate. But since the existing behavior is to fail when an invalid backend is supplied, I wanted to retain that behavior for existing uses and for the use-case herein, support the fallback.

pep517/build.py
jaraco: And the same comment applies here. The `compat_system` provides this functionality.

pep517/envbuild.py
jaraco: It will not. Previously, the output would be emitted unconditionally. Now the output will be redirected to the log, where it can be controlled based on logging settings. I think this library probably needs to step back and specify what callers should expect about output across various operations.

tests/test_meta.py
jaraco: `pep517` is built with [flit](https://pypi.org/project/flit), which is a Python 3 only package. Because these tests are relying on this project's own repo/metadata to run the tests, those tests therefore have the same constraint. It's the same reason you can't install pep517 from source on Python 2.7:

```
$ python2.7 -m pip-run --no-binary=:all: --no-deps pep517                                                                                                                    
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.
Collecting pep517
  Downloading https://files.pythonhosted.org/packages/79/27/1e6a8ba83dab2b243e1020e340552c129b0076fbb5518e0c2a9c44c7d056/pep517-0.5.0.tar.gz
  Installing build dependencies ... error
  Complete output from command /usr/local/opt/python@2/bin/python2.7 /usr/local/lib/python2.7/site-packages/pip install --ignore-installed --no-user --prefix /private/var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T/pip-build-env-jCiSaR/overlay --no-warn-script-location --no-binary :all: --only-binary :none: -i https://pypi.org/simple -- flit:
  DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.
  Collecting flit
    Could not find a version that satisfies the requirement flit (from versions: )
  No matching distribution found for flit
  
  ----------------------------------------
Command "/usr/local/opt/python@2/bin/python2.7 /usr/local/lib/python2.7/site-packages/pip install --ignore-installed --no-user --prefix /private/var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T/pip-build-env-jCiSaR/overlay --no-warn-script-location --no-binary :all: --only-binary :none: -i https://pypi.org/simple -- flit" failed with error code 1 in None
```

These tests could be re-written to use mocks instead of the local pep517 checkout, but I'm satisfied with this approach.

gaborbernat: Should not be that hard to run the tests on a setuptools simple project, not 🤔 flit does generate wheel/sdist that can be installed on Python 2 🤔

pep517/envbuild.py
jaraco: I'm hoping not to have to solve that problem here but only provide a reasonable behavior and avoid emitting the output for API callers.

pep517/build.py
jaraco: I've filed #49 to follow up on this consideration separately.

pep517/envbuild.py
jaraco: This issue is being addressed more generally in #13.

tests/test_meta.py
jaraco: I do think it would be worthwhile for coverage reasons to retrieve metadata from a non PEP 517 source package.

> flit does generate sdist that can be installed on Python 2

Can you demonstrate this behavior? Because the pep517 sdist seems not to have that feature.

gaborbernat: Eh, looking through the CI and tests there's no such tests or coverage report yet. As such for now probably not worth it. Let's release and a subsequent PR can improve the CI to generate coverage, and then we can worry about this.

gaborbernat:
@pfmoore can we land this?

pfmoore:
I'm going to leave that decision to @takluyver as owner of this project. My only real interest is in the wrappers, which pip uses.

pep517/build.py
```diff
@@ -3,25 +3,50 @@
...
 
-@contextlib.contextmanager
-def tempdir():
-    td = tempfile.mkdtemp()
+def validate_system(system):
+    """
+    Ensure build system has the requisite fields.
+    """
+    required = {'requires', 'build-backend'}
+    if required > set(system):
```
takluyver: I don't think this does quite what you expect. Try

```python
if not (required <= set(system)):
```

`>` catches if the keys are a strict subset of the required ones. So it works if you miss one out completely, but if you miss one out and add another - e.g. by making a typo - the keys are no longer a subset.

`<=` checks that the keys are a non-strict superset of the required ones, which is what we care about.

pep517/build.py
```diff
@@ -3,25 +3,50 @@
...
+        missing = required - set(system)
+        message = "Missing required fields: {missing}".format(**locals())
```
takluyver: I'd rather lose the `locals()` introspection trick. It's a distraction if you're not used to reading it, and it doesn't save any characters.

pep517/meta.py
```diff
@@ -0,0 +1,90 @@
...
+
+def build_as_zip(builder=build):
+    with tempdir() as out_dir:
+        builder(dest=out_dir)
+        return dir_to_zipfile(out_dir)
+
+
+def load(root):
+    """
+    Load the metadata from root.
```
takluyver: This could be more descriptive. What is `root`? (I can see from the code that it's a source directory, but it's not obvious from the docstring)

The name also doesn't seem very meaningful - it doesn't load metadata, it builds metadata and then returns a wrapper around it to conveniently access it.

tests/test_meta.py
```diff
@@ -0,0 +1,38 @@
...
+    'sys.version_info < (3,)',
+    reason="pep517 cannot be built on Python 2",
```
takluyver: I'm working towards resolving this, but only in odd scraps of time at the moment.

pep517/build.py
jaraco: I prefer this technique because it most closely translates to the syntax for f-strings, which I'd use here once available. I see the `**locals()` trick as a wart that's a valuable reminder to use f-strings. But since you feel strongly enough, I can omit it.

jaraco:
Thanks for the review. All good points. I'll follow up later.

pep517/build.py
takluyver: Thanks; I understand the point, but I think the downside of the wart outweighs the reminder to use f-strings, which aren't a big win here anyway.

jaraco: Fixed in bdad30a.

pep517/build.py
```diff
@@ -3,25 +3,50 @@
...
+    setuptools.
+    """
     try:
-        yield td
-    finally:
-        shutil.rmtree(td)
+        system = load_system(source_dir)
+    except Exception:
+        system = {}
+    system.setdefault('build-backend', 'setuptools.build_meta')
```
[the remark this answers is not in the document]
jaraco: I can see your point. My instinct was, since pep517 never had the legacy behavior, it's not necessary to drag it into here. That is, the "meta" command will never support that legacy assumption. That means it won't support those projects reliant on that behavior. If I recall, when `__legacy__` was created, it was meant for pip only (and that's why it's `__`-prefixed). I'd prefer to start with this constrained approach and relax the constraint later if demand and consideration indicate.

pep517/build.py
```diff
@@ -3,25 +3,50 @@
...
+    Given a source dir, attempt to get a build system backend
+    and requirements from pyproject.toml. Fallback to
+    setuptools.
+    """
     try:
-        yield td
-    finally:
-        shutil.rmtree(td)
+        system = load_system(source_dir)
+    except Exception:
```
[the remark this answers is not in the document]
jaraco: I do intend to trap both OSError and KeyError here... because the presence of a pyproject.toml without a `build-system` is comparable to no pyproject.toml (for this concern). I'd be tempted to say a malformed pyproject.toml should be treated as having no declared build system, though I do sympathize with your instinct here.

The problem I have with specifying `(OSError, KeyError)` in the `compat_system` function is it couples the internal implementation details of `load_system` with `compat_system`. I guess that's okay as one is meant to wrap the other, but it does make me a little uneasy.

Addressed in 0da9696.

pep517/build.py
jaraco: Good catch. I was able to replicate your findings with this patch:

```
diff --git a/conftest.py b/conftest.py
new file mode 100644
index 0000000..4f1eca8
--- /dev/null
+++ b/conftest.py
@@ -0,0 +1,7 @@
+import sys
+
+collect_ignore = ['pep517/_in_process.py']
+
+
+if sys.version_info < (3,):
+    collect_ignore.append('pep517/build.py')
diff --git a/pep517/build.py b/pep517/build.py
index 781f450..8aafc1c 100644
--- a/pep517/build.py
+++ b/pep517/build.py
@@ -17,6 +17,31 @@ log = logging.getLogger(__name__)
 def validate_system(system):
     """
     Ensure build system has the requisite fields.
+
+    >>> validate_system({'build-backend': None, 'requires': None})
+    >>> validate_system({'build-backend': None, 'requires': None,
+    ...     'other': None})
+
+    >>> validate_system(dict(requires=None))
+    Traceback (most recent call last):
+    ...
+    ValueError: Missing required fields: {'build-backend'}
+
+    >>> validate_system({'build-backend': None})
+    Traceback (most recent call last):
+    ...
+    ValueError: Missing required fields: {'requires'}
+
+    >>> validate_system({})
+    Traceback (most recent call last):
+    ...
+    ValueError: Missing required fields: {'...', '...'}
+
+    >>> validate_system({'build-backend': None, 'other': None})
+    Traceback (most recent call last):
+    ...
+    ValueError: Missing required fields: {'requires'}
+
     """
     required = {'requires', 'build-backend'}
     if required > set(system):
diff --git a/pytest.ini b/pytest.ini
index 5c37966..b5835b8 100644
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,2 +1,2 @@
 [pytest]
-addopts=--flake8
+addopts=--flake8 --doctest-modules
```

(only runs on Python 3)

However, there are too many complications with using doctests on this project, so I'm abandoning that work.

jaraco: Fixed in 1f90e57.

pep517/meta.py
jaraco: Does 7d568e5 clarify?

jaraco:
I believe I've addressed all of the concerns that @takluyver raised. Any remaining concerns or comments?

pep517/build.py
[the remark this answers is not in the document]
takluyver: I suspect we'll soon end up having to use the 'legacy' backend, because the people building a project through this mechanism are not necessarily the ones deciding what assumptions that project makes.

@pfmoore @gaborbernat (and anyone else): any preference on this?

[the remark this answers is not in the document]
gaborbernat: I'd go in +1 to make it setuptools.build_meta:__legacy__ the default.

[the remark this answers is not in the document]
jaraco: Updated in 205d128.

takluyver:
Thanks!

gaborbernat:
@takluyver want to do a release with this? thanks!

takluyver:
There you go, 0.6.0 now out.

takluyver:
PS @jaraco or anyone: if you got a notification about the deploy failure, don't worry. I had forgotten that the releases were automated, and my personal release habit is to upload to PyPI first and then create the tag, so it failed because the packages were already on PyPI.

pep517/build.py
jalanb: @jaraco 

Just passing through, following a link from "GitHub Explore the week", and having decided to concentrate more on doctests (at work): may I ask what complications arose here with using them?

PR #50, opened 2019-05-24
Make zip-safe by relying on importlib.resources

### LINKED ISSUE #6: Wrappers have to be run from a filesystem
Because the `_in_process.py` script is located via `__file__`, the wrappers module won't work from within a zipfile (or other PEP 302 loader). Ideally, that should be fixed (pip runs itself from within a zipfile when run via `get-pip.py`, although it shouldn't be installing anything from source, so it's not clear that it needs PEP 517 support in that scenario).

Just adding this as a reminder for now because I'm not sure what a good solution would be (the wrapper script may be too long to be passed to Python via `-c`).

jaraco:
Fixes #6.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -4,3 +4,4 @@ mock
 testpath
 pytoml
 setuptools>=30
+importlib_resources; python_version < '3.7'
--- a/pep517/wrappers.py
+++ b/pep517/wrappers.py
@@ -214,7 +214,7 @@ class Pep517HookCaller(object):
             # Run the hook in a subprocess
             with resources.path(__package__, '_in_process.py') as script:
                 self._subprocess_runner(
-                    [sys.executable, script, hook_name, td],
+                    [sys.executable, str(script), hook_name, td],
                     cwd=self.source_dir,
                     extra_environ=extra_environ
                 )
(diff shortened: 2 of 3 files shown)


pfmoore:
I'd prefer the solution to #6 to *not* require an additional dependency. As written, this PR would mean that pip (which vendors pep517) would have to vendor `importlib.resources`, for no actual benefit (pip doesn't need to run `pep517` from a zipfile).

Unless there's a better justification for this change than "fixing #6", I'd prefer not to do this.

takluyver:
As I've probably commented elsewhere, my personal view is that the non-filesystem modules ship sailed: there are plenty of modules which already break in one way or another if they're not loaded from regular files and folders. Too many to ever fix. So I don't feel remotely bad using `__file__` in my code.

jaraco:
I believe `importlib.resources` is the best practice for this. You _could_ also use `pkg_resources.resource_filename`, but that code is deprecated in favor of `importlib.resources`, which is part of stdlib since Python 3.7. So you _wouldn't_ need to vendor anything for Python 3.7 and later and you would only need to vendor `importlib_resources` until you drop support for Python 3.6.

If this were my project, I'd want it not to be tied to the same constraints that `pip` has. I'd prefer for this project to present the best practices and then let pip determine how to best incorporate those practices (perhaps vendoring or patching out undesirable functionality).

jaraco:
Let's not waste any more energy on this issue. I'd rather we spend it on the more important metadata issue.

PR #52, opened 2019-06-16
Add a runner argument to Pep517HookCaller

pradyunsg:
Toward #13 

pip doesn't really override this on every invocation, and we're currently patching when initializing anyway. This would be an improvement over status quo.

I'd prefer to discuss the context manager support in a follow up PR.


takluyver:
Thanks, this looks fine in terms of code.

At some point, we should document what's expected of the callable (cmd, cwd & extra_environ args; raise an exception if exit code != 0). This PR seems like a good place to do it, as you're adding it to the docstring anyway. But if you'd rather it was done in a separate PR, that's also fine.

pradyunsg:
Let's do that in a follow up PR.

takluyver:
OK, I've opened #53 for that.

PR #55, opened 2019-07-10
Make pep517 zip-safe when using Python 3.7.

### LINKED ISSUE #6: Wrappers have to be run from a filesystem

duckinator:
**NOTE:** This does not add any new dependencies.

pep517 being zip-safe without an external dependency requires importlib.resources, which first became available in Python 3.7. With older Python versions, this PR falls back to the old behavior.

h/t to @jaraco for #50 which gave me most of the information I needed to figure this out.

Fixes #6.

### DIFF
--- a/pep517/wrappers.py
+++ b/pep517/wrappers.py
@@ -9,6 +9,18 @@ from tempfile import mkdtemp
 
 from . import compat
 
+
+try:
+    import importlib.resources as resources
+
+    def _in_proc_script_path():
+        return resources.path(__package__, '_in_process.py')
+except ImportError:
+    @contextmanager
+    def _in_proc_script_path():
+        yield pjoin(dirname(abspath(__file__)), '_in_process.py')
+
+
 _in_proc_script = pjoin(dirname(abspath(__file__)), '_in_process.py')
 
 
@@ -242,11 +254,12 @@ class Pep517HookCaller(object):
                               indent=2)
 
             # Run the hook in a subprocess
-            self._subprocess_runner(
-                [sys.executable, _in_proc_script, hook_name, td],
-                cwd=self.source_dir,
-                extra_environ=extra_environ
-            )
+            with _in_proc_script_path() as script:
+                self._subprocess_runner(
+                    [sys.executable, str(script), hook_name, td],
+                    cwd=self.source_dir,
+                    extra_environ=extra_environ
+                )
 
             data = compat.read_json(pjoin(td, 'output.json'))
             if data.get('unsupported'):


duckinator:
Also if you have a better/preferred way to track if `importlib.resources` is available than the `_HAVE_IMPORTLIB_RESOURCES` thing, lmk and I'll be happy to change it.

pep517/wrappers.py
```diff
@@ -210,11 +218,19 @@ def _call_hook(self, hook_name, kwargs):
...
 
             # Run the hook in a subprocess
-            self._subprocess_runner(
-                [sys.executable, _in_proc_script, hook_name, td],
-                cwd=self.source_dir,
-                extra_environ=extra_environ
-            )
+            if _HAVE_IMPORTLIB_RESOURCES:
+                with resources.path(__package__, '_in_process.py') as script:
+                    self._subprocess_runner(
```
chrahunt: To avoid duplicating the calls to `_subprocess_runner`, I would define a context manager in each branch of the `try: ... except ImportError: ...` itself that yields the expected path, like

```python
try:
    import importlib.resources as resources

    _in_proc_script_path = lambda: resources.path(__package__, '_in_process.py')
except ImportError:
    from contextlib import contextmanager

    _in_proc_script_path = contextmanager(
        lambda: pjoin(dirname(abspath(__file__)), '_in_process.py')
    )
```

then use `with _in_proc_script_path() as script` unconditionally.

duckinator:
@chrahunt looks like using your suggestion verbatim gives flake8 warnings (for assigning lambda to a variable + >79 characters on a line). About to try a variant of it and hoping 2.7 doesn't explode. :sweat_smile:

duckinator:
@chrahunt okay, it took a few tries (mostly because I only have Py3.7 on this system, so wasn't able to test _everything_ locally), but I think I got it!

jaraco:
Love it. Thanks!

pep517/wrappers.py
```diff
@@ -9,6 +9,18 @@
...
+
+    def _in_proc_script_path():
+        return resources.path(__package__, '_in_process.py')
+except ImportError:
+    @contextmanager
+    def _in_proc_script_path():
+        yield pjoin(dirname(abspath(__file__)), '_in_process.py')
+
+
 _in_proc_script = pjoin(dirname(abspath(__file__)), '_in_process.py')
```
chrahunt: We can get rid of this now. :)

duckinator: @chrahunt handled in #67.

lazka:
From what I see this makes finding `_in_process.py` work, but when it executes it I get

```
  File "/tmp/tmp8w05syzc", line 26, in <module>
    import compat
ModuleNotFoundError: No module named 'compat'
```

Is this expected?

dstufft:
FWIW I can certainly tell from looking here that the current ``_in_process.py`` is not going to work when run from a zipfile, because it's going to extra ``_in_process.py`` to a temporary location. For this to work ``_in_process.py`` is going to need to, as a rule, be self contained OR do some fallback thing like:

```python
try:
    from pep517 import compat
except ImportError:
    import compat
```

However, it's probably a lot simpler and easier to reason about that it just stays self contained.

duckinator:
@lazka @dstufft can you provide more context on your environment, or a testcase? Because looking at it again, yeah, I'd expect problems. But I've been using this in a zipapp as part of a PEP517-compatible build system since January without issue, and now I'm not sure how I was doing so lmao

kitterma:
The context where this came up was in Debian where we unvendor the pip vendored dependencies as wheels (which is supported by pip).  In this case pip is trying to use pep517 as a wheel, running the script and failing.  The details are in https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=954256

There's a debian:unstable docker image you can fire up that makes it trivial to reproduce it (although that's about to change because I just uploaded a change to go back to the vendored pep517 as a work around until this is resolved).

Scott K

lazka:
What @kitterma said (thanks, and thanks for the quick Debian fix!)

@kitterma from what I see you never uploaded python-pip-whl to unstable with an updated pep517, so the error there is the one you get before this PR.

pfmoore:
@kitterma To be fair, while pip supports unvendoring the bundled dependencies, I don't think we've ever suggested that we support debundling them *as wheels*. We wouldn't do that, precisely because the wheel spec [says clearly](https://www.python.org/dev/peps/pep-0427/#is-it-possible-to-import-python-code-directly-from-a-wheel-file) "while this behaviour is a natural consequence of the format design, actually relying on it is generally discouraged".

kitterma:
@pfmoore OK.  The current approach was developed in coordination with pip developers (@dstufft and maybe others, I don't know - it was before I was involved in maintaining pip in Debian).  

Historically we had not debundled pep517 because it was not zip safe (I know now).  So this was started by me trying to use pep517 0.7.0 (the version pip bundles) as a wheel and it not working out well.  Fair enough, distro mistake, distro problem.

The upstream issue is that this update was supposed to make pep517 zip safe, but it didn't.  It's incomplete.  If it had, then updating to the current version would have worked.  Since the additional change needed to achieve the goal of this PR is small, I think it makes sense to go ahead and do it.

I'll be glad to prepare a new request that either inlines compat or adds the fallback import that @dstufft suggested was an option above.  Please let me know which you'd prefer.

dstufft:
@pfmoore That's not true actually, our unbundling logic [explicitly looks for wheels](https://github.com/pypa/pip/blob/master/src/pip/_vendor/__init__.py#L19-L23), and when we still had tests that explicitly tested the unbundled configuration, we were testing them as wheels not as unzipped directories. The process for unbundling is documented [here](https://github.com/pypa/pip/blob/master/src/pip/_vendor/README.rst#debundling), which also explicitly documents the use of wheel files.

We can of course change that, and look for something different, but our unbundling logic is setup the way it is because we don't want distros to just do the simplest thing to unbundle (just unbundle like a normal python package and use the versions installed into ``site-packages``) but instead we want them to effectively re-vendor their patched versions of our dependencies (so that like, ``pip install requests`` still doesn't actually *break* pip to the point you can't use pip anymore). When I worked with barry on this, this was the simplest thing we could get to work that satisfied both of our constraints.

dstufft:
Also this bug doesn't just happen inside wheels, the bug is effectively that pip517 is not zip safe in it's current form. It has some code to make it zip safe, but that code isn't sufficient-- so either pep517 should go the rest of the way and become fully compatible with zip import, or it should rip out the zip import support logic that currently exists and declare zipped imports as unsupported.

A middle ground with some overtures to working under zip import, without *actually* working under zip import seems like the worst of all possible options.

pfmoore:
@dstufft @kitterma My apologies, I was obviously commenting without knowing what I was talking about 🙁 I don't really have a view on how this gets fixed, TBH (or even *if* it gets fixed) - I'll leave the decision to someone who understands the trade-offs better.

duckinator:
Given that the lack of zip safety seems specific to use of `compat.read_json()` and `compat.write_json()`, and how small those functions are, I think inlining it (like @dstufft suggested) is a reasonable approach.

@dstufft @pfmoore would you like me to submit a PR for this?

takluyver:
> It has some code to make it zip safe, but that code isn't sufficient

In so far as people look to me to maintain this project: I don't personally care about zip-safety - I think the idea that you can put arbitrary packages in a zip file and then import them is a ship that sailed many years ago. But I don't object if other people want to make this package zip-safe, so long as what's necessary to do so isn't too drastic.

duckinator:
@takluyver I wouldn't expect it to be too drastic, but if I make a PR for it I'll be sure to @-mention you so you can provide feedback. :+1:

takluyver:
Thanks, go for it :-)

duckinator:
Just opened #78 to fix this.

kitterma:
I've confirmed that it does fix it.  Thanks.

PR #56, opened 2019-08-16
Support back-end path in pyproject.toml

sbidoul:
Hello,

This is a quick shot at supporting `backend-path` in `pyproject.toml`.

All the logic for in-tree backends seems to be in place, only reading the value from `pyproject.toml` was apparently missing.

I made a quick test implementing an in-tree backend that does some pre-processing before delegating to `setuptools.build_meta` and it seems to work fine.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -43,7 +43,11 @@ available:
 
     print(build_sys['requires'])  # List of static requirements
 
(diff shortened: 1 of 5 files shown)


takluyver:
Thanks, this looks good to me.

PR #60, opened 2019-09-14
Raise HookMissing instead of falling back, if requested.

### LINKED ISSUE #58: Option to not fallback for `wrappers.prepare_metadata_for_build_wheel`
Currently `prepare_metadata_for_build_wheel` builds a wheel if the hook is not implemented by the given backend.

It would be nice to be able to disable the automatic fallback behavior. The specific use cases are:

1. More consistent user experience - a package that uses a backend that does not implement `preparing_metadata_for_build_wheel` may show "Preparing wheel metadata" for a long time and "Building wheel for <project> (PEP 517)" for a short time. Another package that does implement `preparing_metadata_for_build_wheel` would show the opposite behavior.
1. When refactoring pip, we would like to align the actual state of the project on disk with the type representing that project in memory. See [pypa/pip#6607 (comment)](https://github.com/pypa/pip/issues/6607#issuecomment-529946244) for reference. Specifically, a "Local modern project" may have its metadata queried but if that fails then we want to know about it and explicitly advance the state to a "Local wheel" by building a wheel. Mocking this behavior for unit tests is very straightforward, but those sorts of tests get further from the truth if the library is keeping a wheel off to the side.

chrahunt:
For now this only applies to `prepare_metadata_for_build_wheel`.

Closes #58.

### DIFF
--- a/pep517/wrappers.py
+++ b/pep517/wrappers.py
@@ -37,6 +37,9 @@ class BackendInvalid(Exception):
 
 class HookMissing(Exception):
     """Will be raised on missing hooks."""
+    def __init__(self, hook_name):
+        super(HookMissing, self).__init__(hook_name)
+        self.hook_name = hook_name
 
 
 class UnsupportedOperation(Exception):
@@ -245,7 +248,7 @@ class Pep517HookCaller(object):
                     message=data.get('backend_error', '')
                 )
             if data.get('hook_missing'):
-                raise HookMissing()
+                raise HookMissing(hook_name)
             return data['return_val']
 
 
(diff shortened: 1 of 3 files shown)


pep517/wrappers.py
```diff
@@ -35,6 +35,10 @@ def __init__(self, backend_name, backend_path, message):
         self.message = message
 
 
+class HookMissing(Exception):
+    """Will be raised on missing hooks."""
```
takluyver: I'd like the exception to carry the name of the missing hook, so that if it's uncaught and reaches the user, it's clear what it was looking for.

takluyver:
Thanks!

PR #62, opened 2019-09-25
Migrate from pytoml to toml.

duckinator:
pytoml is unsupported, and is missing recent features of the TOML standard.

Closes #62.


pfmoore:
Looks good to me. Thanks for the contribution, @duckinator!

PR #63, opened 2019-09-27
Update pyproject.toml

duckinator:
This PR modifies `pyproject.toml` to:

1. Add trove classifiers indicating supported Python versions. This allows automated tools, like PyUp, to verify this information.
2. Update the home-page value, so links are accurate with services like PyPI, PyUp, etc.


takluyver:
Thanks!

PR #66, opened 2019-10-01
Replace temporary subprocess runner on exception

### LINKED ISSUE #65: Default Pep517HookCaller subprocess runner not reverted on exception
If the scope wrapped by [`Pep517HookCaller.subprocess_runner`](https://github.com/pypa/pep517/blob/09b403216f36859c91b554f8488ae217449423c9/pep517/wrappers.py#L130-L137) raises an exception then the overridden `_subprocess_runner` will not get reverted back to the default. We should have a `try: ... finally: ...` around the yield.

chrahunt:
Fixes #65.

### DIFF
--- a/pep517/wrappers.py
+++ b/pep517/wrappers.py
@@ -133,8 +133,10 @@ class Pep517HookCaller(object):
         """
         prev = self._subprocess_runner
         self._subprocess_runner = runner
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks!

PR #70, opened 2019-11-13
Don't pull in importlib_metadata & zipp backports for Python 3.8+

encukou:
The Python 3.8+ standard library `importlib.metadata` and `zipfile.Path` are [preferred over the backports](https://github.com/pypa/pep517/blob/master/pep517/meta.py#L9-L17).

(FWIW: Fedora doesn't have the backports packaged for 3.8, but we use the wheel metadata, so things go [boom!](https://bugzilla.redhat.com/show_bug.cgi?id=1770845) I'm sharing the patch – less dependencies is always good, right?)

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -4,5 +4,5 @@ mock
 testpath
 toml
 setuptools>=30
-importlib_metadata
-zipp
+importlib_metadata ; python_version<'3.8'
+zipp ; python_version<'3.8'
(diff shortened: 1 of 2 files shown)


encukou:
The CI doesn't test 3.8, so I opened https://github.com/pypa/pep517/pull/71 to enable it.
If that is accepted, I can rebase this PR on top and let Travis test it.

takluyver:
Closing & reopening - I believe this will test the merge with master, no need for a rebase.

PR #71, opened 2019-11-13
Update CI to Python 3.8

encukou:
`latest_py3` should now be 3.8 :tada: 

I hope this is the right way to tell it to Travis.


takluyver:
Thanks!

PR #72, opened 2019-11-23
Use flit on Travis CI to publish packages

takluyver:
cc @jaraco - this also removes your (encrypted) credentials from the Travis config, in favour of a scoped token which can upload only this project.


jaraco:
Beautiful. Thanks.

PR #78, opened 2020-04-01
Avoid compat.py, to make _in_process.py fully zip-safe (on Py3.7+).

duckinator:
The relevant part of compat.py is simply included in _in_process.py.

See the recent discussion in https://github.com/pypa/pep517/pull/55#issuecomment-606791307 for context.

cc @lazka @kitterma @dstufft @pfmoore @takluyver

### DIFF
--- a/pep517/_in_process.py
+++ b/pep517/_in_process.py
@@ -14,6 +14,7 @@ Results:
 """
 from glob import glob
 from importlib import import_module
+import json
 import os
 import os.path
 from os.path import join as pjoin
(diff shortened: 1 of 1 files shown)


kitterma:
Thanks for the quick response.  I took the modified version, rebuilt the Debian pep517 package and then used it to rebuild our pip package with pep517 unbundled as a wheel (as discussed in the other thread).  Then I went through the reproducer provided in Debian Bug #954256 [1].

I can confirm that it works correctly now, so this change solves the problem we were having.

@pfmoore Would it be possible to get a new release with this change in it?  If it's not convenient, just let me know and I'll add it as a distro patch until after the next release.  Thanks.

[1] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=954256

pfmoore:
@kitterma Thanks for confirming the fix, I have merged it but I can't do a release myself (it's possible I have the rights, but I don't know the process, and I don't consider myself a release admin on this project). So I'll defer to @takluyver to make a release.

kitterma:
Thanks.

takluyver:
@jaraco did some automation so it should just be a matter of making a tag and watching the wheels turn, though I've been forgetting to use it since then. I'll try to see if it works.

takluyver:
Released as 0.8.2

kitterma:
Excellent.  Thanks.

Issue #82, opened 2020-05-10
Restructure this library to better separate low-level and high-level features


pfmoore:
People seem to regularly expect this library to be a higher-level tool than it is. As a result, we get requests for functionality that we don't want to support, and users are disappointed that features that do exist don't meet their expectations.

I suggest that this library should be reorganised, and potentially split up, to better reflect different features.

1. A very low-level library, that has no external dependencies, that just handles calling backend hooks. Essentially, this would just be the wrappers.py file and support modules.
2. A slightly higher level library that depends on a particular toml library, and parses `pyproject.toml` and calls the backend. It should probably also return the build dependencies. I'm not 100% sure this is useful, but it would satisfy the needs of people wanting to manage dependencies themselves, but still be able to call something that does all the steps of building a project - see for example #81.
3. Maybe a build environment management library - but my feeling is that getting such a thing right is pretty complex. I'd be inclined to say that if we were to provide a library interface to build environment management, it should be based on pip's implementation, *not* on the existing code.
4. A command line tool doing what is currently `pep517.build`, etc. Personally, this isn't something I particularly want to touch, but the growing popularity of `pep517.build` suggests that there's a demand for this.

I'd suggest that this project focus on being a pure library, covering (1) and (2) initially, with (3) a longer-term goal.

The command line functionality can be spun off into an independent project. I'd argue that the current build environment management code should probably go with it - it's not really robust enough to be a usable library at the moment, but it's needed for the command line tools.

pradyunsg:
I'd say #83 solves this. :)

layday:
pep517 is used extensively as a build front-end - whether the authors intended for it or not - for lack of PEP517 front-ends.  What alternative is there following this change?

pfmoore:
https://pypi.org/project/build/

layday:
Thanks.  It appears you have to install from source for the time being since the version on PyPI is not able to handle build requires.

FFY00:
Yes, the current version does not build in isolation. That behavior will change in the new version, it will be enabled by default and you will be able to disable it.

PR #83, opened 2020-06-14
Pull out high-level interfaces as an extended example

takluyver:
An attempt to tackle #82, by putting all the code that tries to manage environments & installation into a separate `buildtool_demo` package, which does not get installed.

This will break things for anyone who was relying on the high-level functionality being in `pep517`, but I don't see any way to clean it up without doing that.

The functions for loading & checking the build-system info are retained as part of the real package, in `pep517.pyproject`.


pradyunsg:
Suggestion: move it into examples/*, to clearly denote that it's an example and make room for us to add additional things if there are any.

takluyver:
OK, I've moved it into an `examples/` folder; I agree that's neater. After some fiddling with CI, the tests embedded in that example are working again.

pradyunsg (commented):
Looks great to me! ^.^

pradyunsg (approved):

takluyver:
Thanks @pradyunsg :-)

pradyunsg:
How do we plan to manage the rollout of this change?

FFY00:
I'd probably wait until a new python-build release/python-build gets into PyPA, so that we can recommend that as an alternative.

pradyunsg:
@FFY00 any ETA on that? anything to help with that?

FFY00:
I need to fix some simple issues and probably improve the test suite a bit. I don't think it should take *too* long.

pganssle:
Am I correct in understanding that this _removes_ `pep517.build`? Was there ever a deprecation warning on this?

I guess I should have taken more seriously the warnings that it wasn't necessarily ready for production use, but given that this was _the only PEP 517 build front-end_ for several years, I would kind of expect it to keep working at least for a while. I'll now have to run around and pin the versions on a bunch of my projects now.

If the justification is "we'll get a release of `python-build` soon, and we'll release this after that", that... doesn't seem like a long enough overlap.

pganssle:
I should note that `pep517.build` is included in this [guide to automatic publishing with Github Actions](https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/) on packaging.python.org.

There's a good chance that it's pretty widely used.

gaborbernat:
IMHO this should be done at shortest 6 months post python-build release to stop hundreds of ci failing all of sudden, and emmit a warning in the meantime about moving to python build. Definitely way to early to proceed with this.

webknjaz:
Also, @jaraco's skeleton relies on this CLI IIRC + I've seen it used in many other places. I use it in projects and I guess since I've put this command into the guide that doesn't help.
So I second the call to have a proper depreciation period + have some effort to clean up the widely used recommendations from tutorials and other examples. While it is not a big deal for me to pin the version, it'd be challenging for folks who are less familiar with packaging.

Julian:
(Another +1 for "this will break CI on every repo I own", so definitely a slowerish release would be super appreciated here too).

takluyver:
If I hold off releasing this change, will everyone stop filing issues about all the shortcomings of `pep517.build`? That's what led to this PR - I wrote the simplest possible PEP 517 frontend as a proof of concept and a starting point for serious frontends, and I failed to make it clear that that's what it was, so everyone's treating this quick sketch like it's meant to be the gold standard.

E.g. a proper build frontend would set up an isolated build environment (using something like venv), and let the user control how build dependencies were satisfied (e.g. get packages from this private index, or ignore the build requirements from the package and install these things instead). It would provide a way to pass config settings to the backend, probably options to control its output. Maybe optimisations for building multiple packages together. `pep517.build` does none of these things!

gaborbernat:
@takluyver I think would be better to make a release that starts emiting that warning instead of just holding off with the release. So users can know they need to migrate in the next (6) months? If you do the release in two months it will still be an abrupt removal without some warning emited beforehand...

takluyver:
Maybe we should hold off on even the warning until python-build is ready? I'm not sure a warning is much use if it can't suggest what to do instead.

Maybe once we can point people to python-build, it won't even be necessary to remove this. To be honest, I hadn't realised just how many people were relying on it. I don't want to break everybody's workflow, I just want to somehow make it clear that the `pep517` package is primarily meant as a component for building PEP 517 frontends (like pip), not a frontend in itself.

gaborbernat:
> Maybe we should hold off on even the warning until python-build is ready?

I think it's ready. The latest release ``0.0.4`` should be good as first release. We don't have to move the project under PyPa to recommend it IMHO (but eventually will move under PyPa... but that's not end user facing change).

webknjaz:
@takluyver that's a good option too. *I think* the major reason `pep517.build` got popular is mainly because pip only has `pip wheel` but there's no command for producing source dists. I *wanted* to use pip for this but couldn't and didn't want to resort to calling setuptools directly.

I didn't realize you were annoyed by `pep517.build` bug reports (I don't think I posted any but still...). May I suggest that as a part of this effort, we could configure a custom issue template chooser? https://docs.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository#configuring-the-template-chooser

There's an example config @ https://github.com/pypa/pip/blob/master/.github/ISSUE_TEMPLATE/config.yml and the preview is here https://github.com/pypa/pip/issues/new/choose.

The idea is that we could add an item that has an external link to `python-build` instead of an issue creation button. This way, you could redirect some of this traffic to the right place. Plus you can add some text description explaining that folks shouldn't create issues but rather switch the tool...

takluyver:
Aha, I hadn't seen the python-build release yet. Thanks @FFY00!

So, how does this plan look:

- [ ] Point to python-build in the README & possibly issue templates (thanks @webknjaz )
- [ ] Make a new branch from before this PR was merged
- [ ] Add a deprecationg warning when `pep517.build` is used
- [ ] Release a version with the warning - possibly after a bit of a wait, to give people a chance to start moving to python-build and flush out any bugs there.

takluyver:
I've made a branch `inc-pep517.build`. I can't spend much time on this right now, but PRs are welcome for warnings & links in the documentation. Other PyPA people with access should feel free to merge stuff if they're confident in it - I don't want to be the bottleneck.

pradyunsg:
Thanks @takluyver! I was reading this and was gonna propose making a branch for the new stuff and I'm glad that you did that already. 🙃

The only thing I'd say is we should have a deprecation warning for *all* of the pep517 CLI stuff, since all of it is going to go away. Off the top if my head, that means also including a warning on pep517.check.

pradyunsg:
#91 exists as the tracking issue and the place for further discussion. :)

PR #95, opened 2020-10-02
Parameterize backend path tests

pganssle:
I went looking through the code a bit to understand a problem in `python-build` and I came across these tests, which had an unused assertion and also seemed like they could do with some parametrization.

### DIFF
--- a/tests/test_inplace_hooks.py
+++ b/tests/test_inplace_hooks.py
@@ -26,7 +26,7 @@ def get_hooks(pkg, backend=None, path=None):
     ['../pkg1', 'subdir/..'],
 ])
 def test_backend_path_within_tree(backend_path):
-    assert Pep517HookCaller(SOURCE_DIR, 'dummy', backend_path)
+    Pep517HookCaller(SOURCE_DIR, 'dummy', backend_path)
 
 
 @pytest.mark.parametrize('backend_path', [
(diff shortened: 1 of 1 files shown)


pradyunsg (approved):

tests/test_inplace_hooks.py
```diff
@@ -20,19 +21,24 @@ def get_hooks(pkg, backend=None, path=None):
...
-def test_backend_path_within_tree():
-    source_dir = pjoin(SAMPLES_DIR, 'pkg1')
-    assert Pep517HookCaller(source_dir, 'dummy', ['.', 'subdir'])
-    assert Pep517HookCaller(source_dir, 'dummy', ['../pkg1', 'subdir/..'])
+@pytest.mark.parametrize('backend_path', [
+    ['.', 'subdir'],
+    ['../pkg1', 'subdir/..'],
+])
+def test_backend_path_within_tree(backend_path):
+    assert Pep517HookCaller(SOURCE_DIR, 'dummy', backend_path)
```
takluyver, line 29:
```suggestion
    Pep517HookCaller(SOURCE_DIR, 'dummy', backend_path)
```

This `assert` doesn't do anything, as far as I can tell. `Pep517HookCaller` doesn't define a `__bool__` method, so it will always pass an assertion if there isn't an exception creating it.

PR #96, opened 2020-10-06
Revert "Pull out high-level interfaces as an extended example"

pradyunsg:
Reverts pypa/pep517#83


pradyunsg:
Merging to unblock `master` from releases and bugfixes.

PR #97, opened 2020-10-13
Allow to specify the isolated build environment

gaborbernat:
Before always defaulted to the current python executable, however if the
isolated build environment was a virtual environment this might not be
the case.

Signed-off-by: Bernát Gábor <bgabor8@bloomberg.net>

Ran into this while doing https://github.com/FFY00/python-build

### DIFF
--- a/pep517/wrappers.py
+++ b/pep517/wrappers.py
@@ -115,8 +115,8 @@ class Pep517HookCaller(object):
     build_backend : The build backend spec, as per PEP 517, from
         pyproject.toml.
     backend_path : The backend path, as per PEP 517, from pyproject.toml.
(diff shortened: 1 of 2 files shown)


pep517/wrappers.py
```diff
@@ -115,6 +115,7 @@ class Pep517HookCaller(object):
     build_backend : The build backend spec, as per PEP 517, from
         pyproject.toml.
     backend_path : The backend path, as per PEP 517, from pyproject.toml.
+    executable: The path to the isolated python executable to run within
     runner : A callable that invokes the wrapper subprocess.
```
pradyunsg, line 119: I'd swap the order here, to match the signature.

PR #103, opened 2021-01-28
tests: remove dependency on mock for python3

eli-schwartz:
Sometimes it was imported as unittest.mock, sometimes as mock. Consistently stick to the former.


eli-schwartz:
Apparently it's using assert_called_once, so gate that to 3.6 instead...

takluyver:
Thanks!

PR #105, opened 2021-01-31
Fix #104 (path pollution)

hexagonrecursion:
Fixes #104

_in_process.py is executed as a subprocess. This causes its parent
directory to become sys.path[0]. Any extra files in the same directory
shadow pypi libraries eg colorlog.

Solution: put _in_process.py in its own special subdirectory.
Note: the directory has to be a module because
importlib.resources.path does not support resources with directory
separators in their name:

https://docs.python.org/3/library/importlib.html#importlib.resources.path :
> resource is the name of the resource to open within package;
> it may not contain path separators and it may not have sub-resources
> (i.e. it cannot be a directory).

PR #106, opened 2021-01-31
Add a test for setup.py support

hexagonrecursion:
It looks like you did not have a test for setup.py support. Now you do. Note: this will unfortunately cause conflicts with #105. After one of the two PRs is merged I'll rebase the other.

### DIFF
--- a/tests/test_call_hooks.py
+++ b/tests/test_call_hooks.py
@@ -171,3 +171,10 @@ def test_issue_104():
         '__init__.py', '__init__.pyc', '_in_process.py', '_in_process.pyc',
         '__pycache__',
     }
+
+
+def test_setup_py():
+    hooks = get_hooks('setup-py')
+    with modified_env({'PYTHONPATH': BUILDSYS_PKGS}):
+        res = hooks.get_requires_for_build_wheel({})
+    assert res == ['wheel']
--- a/tests/test_call_hooks.py
+++ b/tests/test_call_hooks.py
@@ -177,4 +177,5 @@ def test_setup_py():
     hooks = get_hooks('setup-py')
     with modified_env({'PYTHONPATH': BUILDSYS_PKGS}):
         res = hooks.get_requires_for_build_wheel({})
+    res = [x for x in res if x != 'setuptools']
     assert res == ['wheel']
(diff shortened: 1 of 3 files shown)


hexagonrecursion:
Closed due to:
* I do not care about this PR much
* The upstream cares even less (no of feedback for over a month)

tests/test_call_hooks.py
```diff
@@ -151,3 +151,10 @@ def test_custom_python_executable(monkeypatch, tmpdir):
             hooks.get_requires_for_build_wheel()
         runner.assert_called_once()
         assert runner.call_args[0][0][0] == 'some-python'
+
+
+def test_setup_py():
+    hooks = get_hooks('setup-py')
+    with modified_env({'PYTHONPATH': BUILDSYS_PKGS}):
+        res = hooks.get_requires_for_build_wheel({})
+    assert res == ['wheel']  # Shold this be ["setuptools", "wheel"]?
```
takluyver, line 160: I think this is correct. This is asking the backend if it needs any *extra* packages to make a wheel. The backend here is setuptools, and it must already be available, otherwise calling the hook would fail, so it doesn't need to specify itself.

takluyver (commented):
This looks fine if you want to reopen & rebase it.

tests/test_call_hooks.py
takluyver:
```suggestion
    assert res == ['wheel']
```

hexagonrecursion:
Some builds give the below error others pass fine. I'll fix it now.
```
AssertionError: assert ['setuptools', 'wheel'] == ['wheel']
```

hexagonrecursion:
Rebased

tests/test_call_hooks.py
```diff
@@ -171,3 +171,11 @@ def test_issue_104():
         '__init__.py', '__init__.pyc', '_in_process.py', '_in_process.pyc',
         '__pycache__',
     }
+
+
+def test_setup_py():
+    hooks = get_hooks('setup-py')
+    with modified_env({'PYTHONPATH': BUILDSYS_PKGS}):
+        res = hooks.get_requires_for_build_wheel({})
+    res = [x for x in res if x != 'setuptools']
```
takluyver, line 180:
```suggestion
    # Some versions of setuptools list setuptools itself here
    res = [x for x in res if x != 'setuptools']
```

takluyver:
Thanks!

Issue #91, opened 2020-09-09
Moving users away using pep517 as a CLI tool


pradyunsg:
This is the issue for tracking our plans for nudging users to transition away from using pep517 from the CLI. These changes would go into the `inc-pep517.build` branch and *not* the default branch.

- [x] Revert #83 (see #96)
- [x] Point to python-build in the README
- [x] {maybe} Point to python-build from issue templates
- [x] Add a deprecation warning when `pep517.build` is used
- [x] Add a deprecation warning when `pep517.check`is used
- [ ] Release a version with the warning - possibly after a bit of a wait, to give people a chance to start moving to python-build and flush out any bugs there.

_Originally posted by @takluyver in https://github.com/pypa/pep517/pull/83#issuecomment-689525025 and edited by @pradyunsg_

matthewfeickert:
@pradyunsg Apologies in advance is this is covered in an Issue already and I've failed to find it, or if this should be directed to the `build` Issues and not here, but as it isn't mentioned in PR #83, are there plans to also make an [equivalent of `pep517.check` available in the `build` CLI API](https://github.com/scikit-hep/pyhf/issues/1065#issuecomment-692293800)?

webknjaz:
@matthewfeickert I think this question should belong to the python-build tracker. pep517 is going to get rid of the CLI at some point as it wasn't meant to be an official public API. OTOH since python-build is a wrapper around pep517, it's probably reasonable to have check there too.

pganssle:
@pradyunsg What is the goal with this old vs. new branch thing? Do you have any idea what the deprecation period is going to be? I was a little surprised that y'all decided to branch instead of revert.

I will also note that I personally think that while it's a good idea to start incrementally moving people to `python-build`, the current release version is 0.0.4. It needs a _lot_ of love to get to be really stable —  note that it accidentally [doesn't work with `flit` or any other non-`setuptools` projects](https://github.com/FFY00/python-build/issues/98), and [there's a good chance we'll have to change the console script entry point](https://github.com/FFY00/python-build/issues/101).

[...]

PR #108, opened 2021-02-01
Deprecate the high level interface

hexagonrecursion:
#91

> Add a deprecation warning when pep517.build is used
> Add a deprecation warning when pep517.checkis used

### DIFF
--- a/pep517/build.py
+++ b/pep517/build.py
@@ -110,6 +110,8 @@ parser.add_argument(
 
 
 def main(args):
+    log.warning('pep517.build is deprecated. Consider switching to https://pypi.org/project/build/')
+
     # determine which dists to build
     dists = list(filter(None, (
         'sdist' if args.source or not args.binary else None,
--- a/pep517/check.py
+++ b/pep517/check.py
@@ -167,6 +167,8 @@ def check(source_dir):
 
 
 def main(argv=None):
+    log.warning('pep517.check is deprecated. Consider switching to https://pypi.org/project/build/')
+
     ap = argparse.ArgumentParser()
     ap.add_argument(
         'source_dir',


pradyunsg (changes requested):
Thanks for the PR!

I don't think these should be print statements, but rather log.warning(...) calls inside the main function. That'll allow downstream users to use all of logging's capabilities to handle and manage these logs.

pep517/check.py
```diff
@@ -200,4 +200,5 @@ def ansi(s, attr):
 
 
 if __name__ == '__main__':
+    print(ansi('pep517.check is deprecated. Consider switching to https://pypi.org/project/build/', 'red'))
```
pradyunsg, line 203: And, assuming we do change that, this should also not be wrapped in ansi red IMO

hexagonrecursion:
> Thanks for the PR!
> 
> I don't think these should be print statements, but rather log.warning(...) calls inside the main function. That'll allow downstream users to use all of logging's capabilities to handle and manage these logs.

Should I put log.warning in main() or at the top of the module after the imports?

hexagonrecursion:
Done

pep517/build.py
```diff
@@ -110,6 +110,8 @@ def build(source_dir, dist, dest=None, system=None):
 
 
 def main(args):
+    log.warning('pep517.build is deprecated. Consider switching to https://pypi.org/project/build/', stacklevel=2)
```
pradyunsg, line 113: I don't think stacklevel is necessary here.

pradyunsg (commented):
Almost there!

hexagonrecursion:
done

pradyunsg (approved):

PR #112, opened 2021-03-11
Clarify the issue template

hexagonrecursion:
The way the issue template is currently worded someone unfamiliar with the project may think the entire project is deprecated instead of just the CLI tools pep517.build and pep517.check. This PR clarifies the confusion.


takluyver:
Thanks!

PR #119, opened 2021-07-09
Add PEP 660 support

sbidoul:
- [x] tests
- [x] changelog


sbidoul:
This should be ready for review/comments.

It is mostly a straightforward copy of the wheel hooks, with a small difference in the `prepare_metadata_for_build_editable` fallback in case the `build_editable` hook is missing.

Do not merge just yet as I'll do more integration testing with pip.

sbidoul:
I did further testing with pip. This should be good for merging.

takluyver:
Thanks @sbidoul !

PR #122, opened 2021-07-30
avoid tomli Text file object support is deprecated and enable pytest strict mode

graingert:


pyproject.toml
```diff
@@ -20,3 +20,12 @@ classifiers = [
     "Programming Language :: Python :: 3",
 ]
 
+[tool.pytest.ini_options]
+addopts = ["--strict-config", "--strict-markers", "--flake8"]
+xfail_strict = true
+junit_family = "xunit2"
+filterwarnings = [
+    "error",
+    '''ignore:Text file object support is deprecated in favor of binary file objects\. Use `open\("foo.toml", "rb"\)` to open the file in binary mode\.:DeprecationWarning''',
```
graingert, line 29: this deprecation warning is causing downstream failures:
https://github.com/pypa/setuptools/pull/2746/checks?check_run_id=3200086380#step:5:339

pep517/compat.py
```diff
@@ -35,7 +36,15 @@ def read_json(path):
 
 
 if sys.version_info < (3, 6):
-    from toml import load as toml_load  # noqa: F401
+    from toml import load as _toml_load  # noqa: F401
+
+    def toml_load(f):
+        w = io.TextIOWrapper(f, encoding="utf8")
```
graingert, line 42: @hukkin do I need this?
```suggestion
        w = io.TextIOWrapper(f, encoding="utf8", newline="")
```

hukkin: Yes, `encoding="utf8", newline=""` are the correct parameters for opening a TOML file in text mode.

The `load` function from uiri/toml seems like it should support binary file objects though, so I'm not sure that any of this `def toml_load` is needed?

graingert:
```
================================================================================================ FAILURES =================================================================================================
__________________________________________________________________________________ test_missing_backend_gives_exception ___________________________________________________________________________________

    def test_missing_backend_gives_exception():
>       hooks = get_hooks('pkg1')

tests/test_call_hooks.py:40: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/test_call_hooks.py:33: in get_hooks
    data = toml_load(f)
.tox/py35/lib/python3.5/site-packages/toml/decoder.py:156: in load
    return loads(f.read(), _dict, decoder)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

s = b'[build-system]\nrequires = ["eg_buildsys"]\nbuild-backend = "buildsys"\n\n[project]\ndescription = "Factory \xe2\xb8\xbb A code generator \xf0\x9f\x8f\xad"\nmaintainers = [{name = "\xc5\x81ukasz Langa"}]\n'
_dict = <class 'dict'>, decoder = <toml.decoder.TomlDecoder object at 0x7f9c930ef278>

    def loads(s, _dict=dict, decoder=None):
        """Parses string as toml
    
        Args:
            s: String to be parsed
            _dict: (optional) Specifies the class of the returned toml dictionary
    
        Returns:
            Parsed toml file represented as a dictionary
    
        Raises:
            TypeError: When a non-string is passed
            TomlDecodeError: Error while decoding toml
        """
    
        implicitgroups = []
        if decoder is None:
            decoder = TomlDecoder(_dict)
        retval = decoder.get_empty_table()
        currentlevel = retval
        if not isinstance(s, basestring):
>           raise TypeError("Expecting something like a string")
E           TypeError: Expecting something like a string

.tox/py35/lib/python3.5/site-packages/toml/decoder.py:186: TypeError
======================================================================
```

graingert: basestring is a `str` not a `(bytes, str)`

https://github.com/uiri/toml/blob/3f637dba5f68db63d4b30967fedda51c82459471/toml/decoder.py#L14

hukkin: Ah I see, seems like a Python 3 specific bug.

layday:
This won't work with older versions of tomli, will it? Binary file support was added in 1.1.0.

graingert:
> This won't work with older versions of tomli, will it? Binary file support was added in 1.1.0.

Bumped

jaraco:
Sorry to see that toml is going back to Python 1 days. This patch seems good to me and I'd like to see it fixed as it's affecting other projects too.

https://github.com/jaraco/pytest-checkdocs/issues/14

I'm less sure about enabling strict mode.

pytest.ini
```diff
@@ -1,2 +1,10 @@
...
-addopts=--flake8
+addopts =
+    --strict-config
+    --strict-markers
+    --flake8
+xfail_strict = True
+junit_family = xunit2
+filterwarnings =
+    error
+    ignore:SelectableGroups dict interface is deprecated\. Use select\.:DeprecationWarning
```
jaraco, line 10: I'd recommend only filtering this warning for flake8.
```suggestion
    # Suppress deprecation warning in flake8
    ignore:SelectableGroups dict interface is deprecated::flake8
```

jaraco (approved):

pep517/compat.py
takluyver: Given that we always use it with a real file anyway, we could make a compat function that takes a path to a file instead of a file like object. I'm fine with this way if you prefer, though.

graingert: Hey there, I currently can't run CI builds on this project so would prefer to do changes in a followup

takluyver: Sure, you don't have to do it at all - it's just an idea. I'm not sure what you mean about CI, though - it seems to have run fine.

graingert: I've been scrounging build permissions from pypa discord

takluyver: Aha, gotcha.

PR #123, opened 2021-08-01
load json from bytes on py2.7 and py3.6+

graingert:


pep517/compat.py
```diff
@@ -4,28 +4,33 @@
 import sys
 
 
-# Handle reading and writing JSON in UTF-8, on Python 3 and 2.
+# Handle reading and writing JSON in any encoding, on Python 3 and 2.
```
graingert, line 7:
```suggestion
# Handle reading JSON in any encoding and writing JSON in utf8, on Python 3 and 2.
```

takluyver:
Sorry, I don't follow what this is fixing. The JSON used for communication between the frontend and backend processes should *always* be UTF-8, and the compat functions used to write it ensure that. Have you come across a case where it gets written in UTF-16 or something? :confused: 

On Python 3, if you give the JSON module bytes, it just [does some encoding detection and then decodes to str](https://github.com/python/cpython/blob/4d77691172aae81bdcbb0ea75839d0e896c43781/Lib/json/__init__.py#L341). So there shouldn't be a performance benefit of passing it bytes - not that the performance of parsing a small amount of JSON is likely to be an issue here.

graingert:
Mainly it's about making the JSON parsing consistent with toml parsing. Eg passing bytes where possible

takluyver:
OK, but there's no actual reason those things need to be the same, as far as I can see.

PR #124, opened 2021-08-01
load toml using paths

graingert:


pep517/compat.py
```diff
@@ -36,16 +36,21 @@ def read_json(path):
...
             w.detach()
 
     from toml import TomlDecodeError as TOMLDecodeError  # noqa: F401
 else:
-    from tomli import load as toml_load  # noqa: F401
+    from tomli import load as _toml_load_bytes  # noqa: F401
     from tomli import TOMLDecodeError  # noqa: F401
+
+
+def read_toml(path):
```
takluyver, line 54: My thinking when I suggested this was that we would define this function twice *instead of* fiddling with TextIOWrapper. I was thinking of it as a simplification (admittedly a very minor one), rather than an extra layer of abstraction.

takluyver:
This is obsolete now.

PR #125, opened 2021-08-02
compat: fallback to toml if tomli is not available

FFY00:
We already have the code, as we support Python < 3.6, and being able to
fallback to toml if tomli is not available will make bootstraping Python
environments much easier.

Signed-off-by: Filipe Laíns <lains@riseup.net>

### DIFF
--- a/pep517/compat.py
+++ b/pep517/compat.py
@@ -35,7 +35,10 @@ except NameError:
     FileNotFoundError = IOError
 
 
-if sys.version_info < (3, 6):
+try:
+    from tomli import load as toml_load  # noqa: F401
+    from tomli import TOMLDecodeError  # noqa: F401
+except ImportError:
     from toml import load as _toml_load  # noqa: F401
 
     def toml_load(f):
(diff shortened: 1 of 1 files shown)


nanonyme:
I bumped into this as well. Wouldn't it be better to make tomli move to flit instead so it would have setup.py included in its sdist? That way you could bootstrap with sdists using setuptools.

takluyver:
I've actually been working on documenting how to bootstrap things without relying on setuptools: https://github.com/takluyver/flit/pull/441

As things stand, I'd be OK with doing the fallback like this. But I don't want to imply that the recommended way to bootstrap these packages is using setuptools. At some point - possibly quite soon - we'll drop support for Python 3.6, which is already out of security updates from the core developers. When we do, I'd like to go back to just importing one TOML parser (tomli). So bootstrappers should be prepared to do things that way.

FFY00:
Once we get the wheel installer ready, I plan to write a bootstraping script that can build from source, and install, the packages necessary to bootstrap a Python environment, which should fix that issue.

Regardless of the future handling of the TOML parser dependency on this project, simply having a release that works with both `toml` and `tomli` would still be very helpful. So, if you could merge this and tag a release, it would be great. After that, feel free to hard depend on `tomli`.

takluyver:
Closing following discussion on #131 - @FFY00 said this turns out not to be so useful after all.

PR #130, opened 2021-09-06
pep517: in_process: try _in_process.pyc file if _in_process.py doesn't exist

commodo:
This change came about to exist after this bug report:
  https://github.com/openwrt/packages/issues/11912

The general context is that OpenWrt does not ship Python source-code by
default. Instead, the build byte-compiles all .py files into .pyc files and
removes the .py files.
There's a wide reasoning for this, from performance to reducing the size of
these packages.

This change is kept in the Python3 build as this patch:
  https://github.com/openwrt/packages/blob/master/lang/python/python3/patches-pip/001-pep517-pyc-fix.patch

This is an attempt to start a discussion, to see whether it makes sense to
remove it from the OpenWrt packages feed, and live in the pip source-base.

The original author of this patch is Jeffery To.
I am the submitter to this repo.
We're both co-maintainers of the Python3 package in the OpenWrt realm.

Signed-off-by: Jeffery To <jeffery.to@gmail.com>
Signed-off-by: Alexandru Ardelean <ardeleanalex@gmail.com>

### DIFF
--- a/pep517/in_process/__init__.py
+++ b/pep517/in_process/__init__.py
@@ -10,8 +10,13 @@ try:
     import importlib.resources as resources
 
     def _in_proc_script_path():
-        return resources.path(__package__, '_in_process.py')
+        if resources.is_resource(__package__, '_in_process.py'):
+            return resources.path(__package__, '_in_process.py')
+        return resources.path(__package__, '_in_process.pyc')
 except ImportError:
     @contextmanager
     def _in_proc_script_path():
-        yield pjoin(dirname(abspath(__file__)), '_in_process.py')
+        _in_proc_script = pjoin(dirname(abspath(__file__)), '_in_process.py')
+        if not os.path.isfile(_in_proc_script):
+            _in_proc_script = pjoin(dirname(abspath(__file__)), '_in_process.pyc')
+        yield _in_proc_script


commodo:
ping on this :)

pfmoore:
Assuming the `.pyc` is in the same location as the `.py` is wrong. You should probably use `importlib.util.cache_from_source`, but even with that, there could be problems - for example running pip from a metapath hook that references a sourceless pip[^1] - which should be tested.

[^1]: This would happen in real life when setting up an isolated build environment from a sourceless pip, for example.

commodo:
> Assuming the `.pyc` is in the same location as the `.py` is wrong. You should probably use `importlib.util.cache_from_source`, but even with that, there could be problems - for example running pip from a metapath hook that references a sourceless pip[1](#user-content-fn-1-de8c8293481778e25b7a0f5852353604) - which should be tested.
> ## Footnotes
> 
>     1. This would happen in real life when setting up an isolated build environment from a sourceless pip, for example. [leftwards_arrow_with_hook](#user-content-fnref-1-de8c8293481778e25b7a0f5852353604)

phew, it's been a while since i opened this PR;
anyway the use-case for OpenWrt is a bit specific (with regards to .pyc files); we store them along-side with `.py` usually;

i'll re-review this and comeback;

pradyunsg:
Closing this out since its not been updated in a while and there's some significant changes needed for this to even be considered.

Let's keep the discussion in #148 before we get back to this.

Issue #90, opened 2020-09-09
Add type hints to public API


FFY00:
Hey, would it be possible to add type hints to public API? Thanks.

layday:
I added type hints, mostly as an exercise, [here](https://github.com/layday/pep517/commit/79fa35a632d87fcbef9defef6ef27f4b33e3cf97).  I don't know whether this would be a welcome addition so I haven't opened a PR.

takluyver:
I'm open to adding type hints where they're not too ugly. I've never worked with the stub files for Python 2 compatible hints, though, so it's entirely possible I'd just overlook them when things changed.

layday:
Would you like me to open a PR for the commit above?

xavfernandez:
I don't know what pep517 is planning but pip will be dropping python 2 support in January 2021. If pep517 plans to do the same, it might want to wait 2,5 months to add type hints inline vs the stub route?

FFY00:
Why don't we use type hint comments? They are Python 2 compatible and even if we drop Python 2 support we wil have to use them if we don't plan on dropping Python 3.5.

layday:
Why would you need to use comments in 3.5?  As for placing type hints in-line, you'd either have to conditionalise importing `typing` (how to do this [isn't always obvious](https://github.com/pypa/packaging/blob/61672bf/packaging/_typing.py)) and quote annotations, or require `typing` as a runtime dependency.

takluyver:
As a boring, low-level piece of infrastructure I'd expect `pep517` to be generally quite slow to drop old Python versions. If the spec changes (like it did for `backend-path`), the library will need to change. And tools using the library probably want the spec to be covered on all versions of Python those tools still support.

That said, I think pip is also pretty conservative, so unless someone comes to say they need Python 2 support for another few months, I'm OK with dropping it not long after pip does.

jdufresne:
> That said, I think pip is also pretty conservative, so unless someone comes to say they need Python 2 support for another few months, I'm OK with dropping it not long after pip does.

I started this process in https://github.com/pypa/pep517/pull/131. I think once there is no Python 2 compatibility, there will be fewer hurdles to add type hints.

pradyunsg:
Once #151 lands, there will be a significantly smaller API surface as well.

pradyunsg:
[...]

PR #131, opened 2021-09-06
Remove support for end-of-life Pythons

jdufresne:
Match pip support Pythons and remove support for end-of-life Pythons.
pip has not support Python 2 since 21.0 (2021-01-23). For dates on when
these environments went EOL, see:

https://devguide.python.org/devcycle/#end-of-life-branches

Removing support for Python2 allows for several simplifications:

- Remove use of toml package in favor of tomli
- Remove compat.py
- Replace io.open with builtin open
- Replace tempdir function with tempfile.TemporaryDirectory
- Replace mkdir_p function with os.makedirs
- Remove use of mock package in favor of unittest.mock
- Remove import from future module
- Remove unnecessary inheritance from object

Refs #90


hugovk:
Here's the pip installs for pep517 from PyPI for October 2021 showing low numbers for EOL versions (<3.6):

| category | percent | downloads |
|:---------|--------:|----------:|
| 3.8      |  67.63% | 4,113,564 |
| 3.9      |  16.02% |   974,421 |
| 3.7      |   9.80% |   596,365 |
| 3.6      |   4.26% |   259,225 |
| 3.10     |   1.58% |    96,366 |
| null     |   0.42% |    25,371 |
| 2.7      |   0.22% |    13,406 |
| 3.5      |   0.05% |     3,287 |
| 3.11     |   0.01% |       402 |
| 3.4      |   0.00% |        24 |
| 3.3      |   0.00% |         5 |
| Total    |         | 6,082,436 |

Date range: 2021-10-01 - 2021-10-31

Source: `pip install -U pypistats && pypistats python_minor pep517 --last-month`

pep517/build.py
```diff
@@ -1,15 +1,14 @@
 """Build a project using PEP 517 hooks.
 """
 import argparse
-import io
 import logging
 import os
 import shutil
+import tomli
```
takluyver, line 7: tomli is a third-party module, so it should go in a separate group, after the stdlib modules and before the modules that are part of this package. I'm aware I don't always follow this convention very strictly itself, but I think it's a good opportunity to follow it here.

(This can be cleaned up in a few other files as well - I won't comment on each one)

pyproject.toml
```diff
@@ -9,13 +9,17 @@ author-email = "thomas@kluyver.me.uk"
...
+requires-python = ">=3.6"
 classifiers = [
     "License :: OSI Approved :: MIT License",
-    "Programming Language :: Python :: 2.7",
     "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: 3.6",
+    "Programming Language :: Python :: 3.7",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
```
takluyver, line 24:
```suggestion
```

I deliberately leave these classifiers out. The `requires-python` specifies the same thing in a more precise and useful way (useful because pip respects it), and doesn't have to be updated when a new Python version comes out.

takluyver:
Thanks, and sorry it's taken me a while to get round to looking at this. It basically looks good, and the statistics @hugovk shows make a good case for it.

@FFY00: do you still want a release which falls back to loading `toml` even on newer versions of Python, as in #125? This PR removes the use of toml altogether, so if you still need that, we can hold off merging this for a little bit longer.

FFY00:
I don't think we need it. In `build`, it turns out, we don't actually use anything that requires loading TOML, so it is possible to bootstrap without it anyway.

takluyver:
Great, in that case I'll close #125 and go ahead with this simplification. :slightly_smiling_face:

pep517/build.py
jdufresne: :+1: Will fix. 

I recommend adopting a tool such as isort to do this step automatically. I think it will make contributing and reviewing simpler.

I've added isort in #137.

pyproject.toml
jdufresne: No problem. I will remove it.

FWIW, some users use this information to determine the suitability of library within a particular project. It helps communicate  precise minor versions support and how up to date a release is. Further, the trove classifiers are easier to parse so it is useful for scripts.

jdufresne:
Thanks for the review @takluyver ! I have addressed all feedback in the latest revision.

takluyver:
Thanks @jdufresne !

PR #133, opened 2021-10-13
Add _supported_features()

sbidoul:
Pip needs a mechanism to detect if build_editable is supported, in order to properly support legacy editable installs for projects that also have a pyproject.toml.

We provide the information with a `_supported_hooks` hook.

Other frontends should normally not need this. Should it be the case, a proper mechanism to expose backend capabilities should standardized.

See https://github.com/pypa/pip/pull/10577#issuecomment-941222164 for more information.

### DIFF
--- a/pep517/in_process/_in_process.py
+++ b/pep517/in_process/_in_process.py
@@ -103,6 +103,19 @@ def _build_backend():
     return obj
 
 
+def _supported_features():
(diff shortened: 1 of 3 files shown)


sbidoul:
Not sure why this is red. Did I manage to unearth a flake8 bug ?

pradyunsg:
@takluyver nudge.

PR #137, opened 2021-11-25
Use isort to automatically sort Python imports

jdufresne:
isort is a Python utility to sort imports alphabetically by section. It
improves the contributor experience by catching style concerns early in
CI rather than during a review. When CI fails, the contributor can use
the tool to apply changes automatically.

isort is very popular and used by big Python projects such as pip.

https://pycqa.github.io/isort/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,3 +19,6 @@ classifiers = [
     "Programming Language :: Python :: 2.7",
     "Programming Language :: Python :: 3",
 ]
+
+[tool.isort]
+profile = "black"
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,6 +3,23 @@ name: Tests
 on: [push, pull_request]
 
 jobs:
+  isort:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: Set up Python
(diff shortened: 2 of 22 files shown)


pyproject.toml
```diff
@@ -19,3 +19,6 @@ classifiers = [
     "Programming Language :: Python :: 2.7",
     "Programming Language :: Python :: 3",
 ]
+
+[tool.isort]
+profile = "black"
```
pradyunsg, line 24: You also need to set the multiline output line mode.

https://pycqa.github.io/isort/docs/configuration/black_compatibility.html#using-a-config-file-such-as-isortcfg

pradyunsg (dismissed):

pyproject.toml
jdufresne: According to the isort docs here: https://pycqa.github.io/isort/docs/configuration/profiles.html

That is `multi_line_output = 3` is included in the black profile. Did you mean something else?

jdufresne: Confirmed with the isort author that adding `multi_line_output` is unnecessary with the profile:

https://github.com/PyCQA/isort/pull/1858

pradyunsg: Neato! Thanks for doing this. I was just looking to make the same PR. :)

pradyunsg (approved):

PR #139, opened 2022-01-10
Switch build backend from flit to setuptools for bootstrapping.

### LINKED ISSUE #138: pep517 requiring PEP-517 support to build creates a circular dependency for build frontends depending on pep517
As [pointed out](https://github.com/pypa/build/issues/430#issuecomment-1008662604) by @layday [`pep517`](https://github.com/pypa/pep517/) currently requires a [PEP-517](https://www.python.org/dev/peps/pep-0517/) compatible build frontend to build, however this means no [PEP-517](https://www.python.org/dev/peps/pep-0517/) build frontends like [`build`](https://github.com/pypa/build/) may depend on [`pep517`](https://github.com/pypa/pep517/) without introducing a circular dependency chain.

One potential way of solving this issue is using setuptools along the lines of [`build`](https://github.com/pypa/build) which supports a simple [`setup.py` shim](https://github.com/pypa/build/blob/0.7.0/setup.py) for bootstrapping to break the dependency cycle.

jameshilliard:
Fixes #138.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,24 +1,6 @@
 [build-system]
-requires = ["flit_core >=2,<4"]
-build-backend = "flit_core.buildapi"
-
-[tool.flit.metadata]
-module = "pep517"
-author = "Thomas Kluyver"
-author-email = "thomas@kluyver.me.uk"
-home-page = "https://github.com/pypa/pep517"
-description-file = "README.rst"
-requires = [
-    "tomli >=1.1.0",
-    "importlib_metadata;python_version<'3.8'",
-    "zipp;python_version<'3.8'",
-]
-requires-python = ">=3.6"
-classifiers = [
-    "License :: OSI Approved :: MIT License",
-    "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3 :: Only",
-]
+requires = ["setuptools >= 40.8.0", "wheel"]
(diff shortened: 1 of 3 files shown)


takluyver:
:-1:

pfmoore:
-1 from me as well. Setuptools must not be the only option for bootstrapping the ecosystem, or we're stuck in the same situation of one backend being special that we're trying to avoid with our standardisation efforts.

jameshilliard:
> Setuptools must not be the only option for bootstrapping the ecosystem

Multiple options is fine but I don't see why setuptools shouldn't be an option at all for a critical toolchain dependency like this, I mean it's by far the best positioned for pep517 bootstrapping in general since everyone already has useable build+install tooling for it already.

> we're stuck in the same situation of one backend being special that we're trying to avoid with our standardisation efforts

So it should be `flit` instead? I mean that sounds to me like we would just be swapping one special backend for another.

Anyways distributions don't really want many different build backends in the first place(since that's just more maintenance burden for us but at least it's manageable as long as there aren't circular dependencies) but having multiple different frontends required just for pep517 bootstrapping is a problem as it really makes the integration a good bit more complex.

takluyver:
Closing this for now, so we can try to keep discussion in one place: https://github.com/pypa/build/issues/430

We can reopen it later if necessary, but I don't think that's likely.

PR #145, opened 2022-06-15
Propagate 'message' in BackendInvalid

aws-taylor:
Propagate the 'message' field up to the parent Exception class so that errors are displayed more clearly on the command line. Prior to this change, a user sees the following when a BackendInvalid exception is thrown by a build backend:

```
> python3 -m pip install <some package>
<traceback>
pip._vendor.pep517.wrappers.BackendInvalid
```

And after, the user will see:

```
python3 -m pip install <some package>
<traceback>
pip._vendor.pep517.wrappers.BackendInvalid: <message>
```


takluyver:
Thanks!

PR #147, opened 2022-08-02
Run flake8 static checks separately from test suite

takluyver:
pytest-flake8 is a compatibility headache - see https://github.com/tholo/pytest-flake8/issues/81 https://github.com/tholo/pytest-flake8/issues/83 & https://github.com/tholo/pytest-flake8/issues/87 . The advantages of running it inside pytest seem small at best, so this gets rid of the pytest integration in favour of running `flake8` directly. In the CI, isort & flake8 are run as part of a 'static-checks' job.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,7 +6,7 @@ env:
   FORCE_COLOR: 1
 
 jobs:
-  isort:
+  static-checks:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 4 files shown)


takluyver:
To be clear, I'm not trying to criticise the developers of pytest-flake8 - I think compatibility headaches are probably unavoidable. Flake8 is pretty clear that it's an application, not a library, and plugin APIs (it's a pytest plugin) are always difficult, because eventually the application developers are going to have a choice between slightly breaking a plugin API and holding back the application development.

I'm going to merge this since one of those compatibility issues is currently causing the CI to fail.

PR #150, opened 2022-10-17
Use tomllib on Python 3.11+ also in tests and docs

hroncok:
I assume the README is doc-tested, so I've changed that one as well.


README.rst
```diff
@@ -22,12 +22,17 @@ Usage—you are responsible for ensuring build requirements are available:
 .. code-block:: python
 
     import os
-    import tomli
+    try:
+        import tomllib
+    except ImportError:
+        import tomli as tomllib
```
takluyver, line 28:
```suggestion
    try:
        import tomllib  # Python >= 3.11 - standard library
    except ImportError:
        import tomli as tomllib  # Python <= 3.10
```

Just to make this clear to people

PR #151, opened 2022-10-29
Rename the project to `pyproject-hooks` and trim scope

### LINKED ISSUE #49: Build/check should fallback to setuptools unconditionally
In [this comment](https://github.com/pypa/pep517/pull/48#discussion_r287282963), @gaborbernat suggests that the project should unconditionally fallback to setuptools if PEP 517 is not indicated.

That seems like a plausible suggestion, but I'd like to separate that concern from the concern in the PR.

### LINKED ISSUE #91: Moving users away using pep517 as a CLI tool

### LINKED ISSUE #136: Rename `pep517` to `pyproject-hooks`, and trim functionality
This could be somewhat disruptive, but with #119, this project's name is objectively incorrect.

Given that this was originally written as a sample implementation of the standard and has since become the canonical place to provide wrappers/callers into pyproject.toml-based Python package builds, it might make sense to completely deprecate this package (not just the `pep517.build/check` interfaces, as discussed in #91) and move the wrappers into a separate package.

Let's switch to a more user-friendly name here, that better conveys the purpose of this package?

### LINKED ISSUE #142: Silencing stdout/stderr by default makes some errors extremely hard to debug
Using `pep517` to extract the project version number resulted in the following error:

```
  File "/usr/local/bin/get-version", line 7, in <module>
    meta = pep517.meta.load(".")
  File "/usr/local/lib/python3.9/site-packages/pep517/meta.py", line 71, in load
    path = Path(build_as_zip(builder))
  File "/usr/local/lib/python3.9/site-packages/pep517/meta.py", line 58, in build_as_zip
    builder(dest=out_dir)
  File "/usr/local/lib/python3.9/site-packages/pep517/meta.py", line 53, in build
    _prep_meta(hooks, env, dest)
  File "/usr/local/lib/python3.9/site-packages/pep517/meta.py", line 28, in _prep_meta
    reqs = hooks.get_requires_for_build_wheel({})
  File "/usr/local/lib/python3.9/site-packages/pep517/wrappers.py", line 179, in get_requires_for_build_wheel
    return self._call_hook('get_requires_for_build_wheel', {
  File "/usr/local/lib/python3.9/site-packages/pep517/wrappers.py", line 329, in _call_hook
    self._subprocess_runner(
  File "/usr/local/lib/python3.9/site-packages/pep517/wrappers.py", line 76, in quiet_subprocess_runner
    check_output(cmd, cwd=cwd, env=env, stderr=STDOUT)
  File "/usr/local/lib/python3.9/subprocess.py", line 424, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/local/lib/python3.9/subprocess.py", line 528, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['/usr/local/bin/python3', '/usr/local/lib/python3.9/site-packages/pep517/in_process/_in_process.py', 'get_requires_for_build_wheel', '/tmp/tmp5zbzjtns']' returned non-zero exit status 1.
```

Note that the line-numbers may not be representative as I had to resort to some fairly ugly print-debugging with a pinch of `pdb`.

After a lengthy session I found [quiet_subprocess_runner](https://github.com/pypa/pep517/blob/a942316305395f8f757f210e2b16f738af73f8b8/pep517/wrappers.py#L70) which completely suppresses any output. Wrapping that with a try/except and *printing* the exception content has shown me this message:

```
b'0\n1\n2\nrunning egg_info\nerror: [Errno 13] Permission denied\n'
```

Which is pretty clear and helped me to fix the issue in my environment.

But this should not have taken me over an hour to fix. At *least* put a debug-log or something with the error to make it easier to detect. My first try was actually to set the root-logger level to `DEBUG` but I could also not see anything there.

pradyunsg:
Toward #136 
Closes #142, by removing the functionality it is requesting changes to.
Closes #91, by removing the functionality that it is tracking removal of.
Closes #49, by removing the functionality it is requesting changes to.

This is the first step, listed in https://github.com/pypa/pep517/issues/136#issuecomment-1229152927.


pradyunsg:
There are a few more "obvious" improvements, like moving to `[project]` metadata, automated formatting, improving the documentation and adding type annotations.

I suggest that we make those improvements in follow ups.

src/pyproject_hooks/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+    BackendInvalid,
+    BackendUnavailable,
+    BuildBackendHookCaller,
+    HookMissing,
+    UnsupportedOperation,
+    default_subprocess_runner,
+    quiet_subprocess_runner,
+)
+
+__version__ = '0.13.0'
```
Secrus, line 14: Wouldn't this call for either a major version bump or versioning reset, since this practically makes this a new project from an end-user perspective?

README.rst
```diff
@@ -23,7 +23,7 @@ Usage—you are responsible for ensuring build requirements are available:
 
     import os
     import tomli
-    from pep517.wrappers import Pep517HookCaller
+    from pyproject_hooks.wrappers import BuildBackendHookCaller
```
takluyver, line 26:
```suggestion
    from pyproject_hooks import BuildBackendHookCaller
```

src/pyproject_hooks/__init__.py
takluyver: I don't think it's required, because with the name change there's no risk of unwittingly upgrading. But I'd be happy enough to call it 1.0 now that we're cutting off the deprecated bits of API.

takluyver:
Thanks @pradyunsg !

PR #153, opened 2022-10-31
Remove unused/duplicate import from example in README

michael-k:
`tomli` is already imported as `tomllib` on Python <= 3.10


takluyver:
Thanks for spotting this!

PR #152, opened 2022-10-31
Various cleanup and prepare for 1.0 release

takluyver:
Updating docs, package metadata and dependencies.

PR #154, opened 2022-11-03
Rework the documentation based on the renaming of the project

pradyunsg:
Toward #136 

@takluyver I just noticed #152 (🤦🏽) and this has a few overlapping changes with that.


README.rst
```diff
@@ -1,53 +1,8 @@
...
-
-    config_options = {}   # Optional parameters for backend
-    # List of dynamic requirements:
-    print(hooks.get_requires_for_build_wheel(config_options))
-    # Again, the caller is responsible for installing these build requirements
-
-    destination = 'also/a/folder'
-    whl_filename = hooks.build_wheel(destination, config_options)
-    assert os.path.isfile(os.path.join(destination, whl_filename))
+You can read more in the `documentation <https://pyproject-hooks.readthedocs.org/>`.
```
takluyver, line 8:
```suggestion
You can read more in the `documentation <https://pyproject-hooks.readthedocs.io/>`.
```

doc/Makefile
```diff
@@ -1,20 +0,0 @@
-# Minimal makefile for Sphinx documentation
```
takluyver, line 1: Is there a reason to drop the Makefile? `make html` is how I'm used to building Sphinx docs locally - am I out of date?

docs/pyproject_hooks.rst
```diff
@@ -0,0 +1,45 @@
...
+
+Exceptions
+----------
+
+Each exception has public attributes with the same name as their constructors.
+
+.. autoclass:: pyproject_hooks.BackendInvalid
+.. autoclass:: pyproject_hooks.BackendUnavailable
+.. autoclass:: pyproject_hooks.HookMissing
+.. autoclass:: pyproject_hooks.UnsupportedOperation
```
takluyver, line 45:
```suggestion
.. autoexception:: pyproject_hooks.BackendInvalid
.. autoexception:: pyproject_hooks.BackendUnavailable
.. autoexception:: pyproject_hooks.HookMissing
.. autoexception:: pyproject_hooks.UnsupportedOperation
```

Maybe? Not a big difference, but since autoexception exists...

docs/release-process.rst
```diff
@@ -0,0 +1,19 @@
...
+    pip install bump2version
+
+- Run it to bump the version strings within the project, like::
+
+    bump2version minor
+
+- Push the commit and tags to GitHub.
+- To build and upload the distribution files, run::
+
+    flit publish
```
takluyver, line 19:
```suggestion
  A GitHub action will upload the release to PyPI.
```

(Though I'll need to do one manually after the rename and then update the token on Github)

takluyver:
Thanks! No worries about the conflict with #152; I'll fix them once this is in.

doc/Makefile
pradyunsg: For the last question: Maybe? :P 

I tend to run `sphinx-build docs build/docs`, which I find straightforward-enough. I am down to add a tox target for the documentation build instead -- since realistically, there's both environment management *and* the documentation build command that are reasonably coupled. I'm ambivalent on the `make html` but if we do want to keep those, I'd prefer to file a follow up adding those back, as generated by the latest Sphinx version.

pradyunsg:
@takluyver A gentle nudge for a re-review. :)

pradyunsg:
On the basis that:

- we can address any potential feedback around the documentation in a follow up
- things around reaching the `pypa/pep517` are a bit confusing without this change
- this is a significant-enough improvement to the docs overall
- it's easier to ask for forgiveness rather than permission

I'm gonna go ahead and merge this without waiting on a re-review. If someone reckons that might be overstepping or would otherwise prefer that I wait in the future, please let me know!

takluyver:
Thanks Pradyun, and sorry I didn't get around to reviewing it again - I'm happy, in any case.

PR #155, opened 2022-11-03
Add warnings on all documentation about `pep517` being unsupported

pradyunsg:
Toward #136 

This serves to inform users to *not* use `pep517`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,7 +3,10 @@ API to call PEP 517 hooks
 
 .. warning::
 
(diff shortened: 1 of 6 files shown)


takluyver:
I'd suggest slightly different wording to reassure people that the (main) API still exists, just at a different name. Something like:

> The core of this package has been renamed to [pyproject-hooks](https://pyproject-hooks.readthedocs.io/en/latest/). Please use that package (low level) or [build](https://pypa-build.readthedocs.io/) (higher level) in place of `pep517`.

pradyunsg:
Works for me!

takluyver:
Thanks!

PR #159, opened 2022-11-21
Bump version: 0.13.0 → 1.0.0

takluyver:
As a PR, since I can no longer commit to master.


takluyver:
I did, but I had to do the first upload manually before I could create a suitable token, so the automated upload failed. It should work for the next release.

takluyver:
(Or I thought I had to - I hadn't realised at the time that there was a 0.1 release already there from 2021)

PR #163, opened 2023-02-06
Modernise the codebase with black, ruff, pre-commit and nox

pradyunsg:
This mostly exists to make it easier for me to work on this codebase, by using tools that aid the workflow by handling formatting and automating the workflow more gracefully than the existing configured tools (tox, flake8).

Closes #162 since this includes that.


takluyver:
Thanks Pradyun!

PR #166, opened 2023-04-03
Rename various pieces to match the new project name

pradyunsg:
This aligns various existing spots that used the old project name to use either the new project name or a name that more clearly reflects what the piece does.


takluyver:
Thanks!

PR #167, opened 2023-04-05
Type annotate the public API

### LINKED ISSUE #90: Add type hints to public API

pradyunsg:
Closes #90


src/pyproject_hooks/_impl.py
```diff
@@ -39,19 +45,35 @@ def __init__(self, backend_name, backend_path, message):
...
 
-def default_subprocess_runner(cmd, cwd=None, extra_environ=None):
+class SubprocessRunner(Protocol):
+    """A protocol for the subprocess runner."""
+
+    def __call__(
+        self,
+        cmd: List[str],
+        cwd: Optional[str] = None,
+        extra_environ: Optional[Dict[str, str]] = None,
```
layday, line 67: It's considered [good practice](https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#wide-vs-narrow-types) to use collection ABCs for function parameters:

```suggestion
        cmd: Sequence[str],
        cwd: Optional[str] = None,
        extra_environ: Optional[Mapping[str, str]] = None,
```

src/pyproject_hooks/_impl.py
```diff
@@ -104,12 +130,12 @@ class BuildBackendHookCaller:
...
-        source_dir,
-        build_backend,
-        backend_path=None,
-        runner=None,
-        python_executable=None,
-    ):
+        source_dir: str,
+        build_backend: str,
+        backend_path: Optional[List[str]] = None,
+        runner: Optional["SubprocessRunner"] = None,
```
layday, line 136: I don't think the quotes here are necessary:

```suggestion
        runner: Optional[SubprocessRunner] = None,
```

src/pyproject_hooks/_impl.py
```diff
@@ -132,7 +158,7 @@ def __init__(
         self.python_executable = python_executable
 
     @contextmanager
-    def subprocess_runner(self, runner):
+    def subprocess_runner(self, runner: "SubprocessRunner") -> Iterator[None]:
```
layday, line 161:
```suggestion
    def subprocess_runner(self, runner: SubprocessRunner) -> Iterator[None]:
```

src/pyproject_hooks/_impl.py
```diff
@@ -169,8 +198,11 @@ def get_requires_for_build_wheel(self, config_settings=None):
         )
 
     def prepare_metadata_for_build_wheel(
-        self, metadata_directory, config_settings=None, _allow_fallback=True
-    ):
+        self,
+        metadata_directory: str,
+        config_settings: Optional[Dict[str, Any]] = None,
+        _allow_fallback: bool = True,
+    ) -> Optional[str]:
```
layday, line 205: The docstring's `rtype` is a `str`, should be updated if it can also return `None`.  IIUC, this will only return `None` if `_allow_fallback` is `True`, so an overload might be more appropriate.

src/pyproject_hooks/_impl.py
```diff
@@ -6,31 +6,37 @@
...
     with open(path, encoding="utf-8") as f:
         return json.load(f)
 
 
 class BackendUnavailable(Exception):
     """Will be raised if the backend cannot be imported in the hook process."""
 
-    def __init__(self, traceback):
+    def __init__(self, traceback: str) -> None:
         self.traceback = traceback
```
layday, line 28: Public attributes need to be annotated:

```suggestion
        self.traceback: str = traceback
```

layday (commented):
Should have `py.typed` marker file inside the `pyproject_hooks` package.

takluyver:
Thanks @pradyunsg - it looks like @layday has already done a detailed review of the annotations, so I'll skip doing that.

My inclination is to be fairly conservative with Python version support, since this is meant to be a boring piece of low-level infrastructure that doesn't change often. I see pip hasn't dropped 3.7 support yet, so I'd suggest we don't here either.  But I'm not going to argue about this if you really want to.

src/pyproject_hooks/_impl.py
pradyunsg: Done.

src/pyproject_hooks/_impl.py
pradyunsg: It isn't for 3.8+, but we're preserving 3.7 compatibility (see below).

src/pyproject_hooks/_impl.py
pradyunsg: It isn't for 3.8+, but we're preserving 3.7 compatibility (see below).

takluyver (approved):

layday (approved):
I've type-checked this branch against the main branch of build; there was only one error, which looks like something that should be fixed on build's side:

```
src/build/__init__.py:180: error: Argument "runner" to "BuildBackendHookCaller" has incompatible type "Callable[[Sequence[str], Optional[str], Optional[Mapping[str, str]]], None]"; expected "Optional[SubprocessRunner]"  [arg-type]
```

(This is presumably because the protocol has default arguments.)

layday:
https://github.com/pypa/build/pull/601

pradyunsg:
The latest commit makes it possible to do...

```
if TYPE_CHECKING:
    from pyproject_hooks import SubprocessRunner
```

And then annotate subprocess runners using that type.

pradyunsg:
Instead of leaving that as a comment, I've gone ahead and documented that.

src/pyproject_hooks/_impl.py
henryiii: This shouldn’t be needed, type checkers infer this from the arguments if it is `__init__`.

henryiii:
Wouldn’t it make sense to allow `SubprocessRunner` at runtime on Python 3.8+? So this interface oddity will go away once 3.7 is dropped? Another idea (if it’s not runtime checkable) is to just make it a dummy object at runtime, so downstream code isn’t forced into extra TYPE_CHECKING checks.

layday:
That's what we do with the `IsolatedEnv`, it's a simple `ABC` at runtime on Python 3.7.

src/pyproject_hooks/_impl.py
```diff
@@ -39,19 +59,23 @@ def __init__(self, backend_name, backend_path, message):
...
 
 
-def default_subprocess_runner(cmd, cwd=None, extra_environ=None):
+def default_subprocess_runner(
+    cmd: Sequence[str],
+    cwd: Optional[str] = None,
+    extra_environ: Optional[Mapping[str, str]] = None,
```
gotmax23, line 77:
```suggestion
    extra_environ: Mapping[str, str] | None = None,
```

src/pyproject_hooks/_impl.py
```diff
@@ -39,19 +59,23 @@ def __init__(self, backend_name, backend_path, message):
...
 
-    def __init__(self, traceback):
+    def __init__(self, traceback: str) -> None:
         self.traceback = traceback
 
 
-def default_subprocess_runner(cmd, cwd=None, extra_environ=None):
+def default_subprocess_runner(
+    cmd: Sequence[str],
+    cwd: Optional[str] = None,
```
gotmax23, line 76: See above.

src/pyproject_hooks/_impl.py
```diff
@@ -63,7 +87,11 @@ def default_subprocess_runner(cmd, cwd=None, extra_environ=None):
     check_call(cmd, cwd=cwd, env=env)
 
 
-def quiet_subprocess_runner(cmd, cwd=None, extra_environ=None):
+def quiet_subprocess_runner(
+    cmd: Sequence[str],
+    cwd: Optional[str] = None,
+    extra_environ: Optional[Mapping[str, str]] = None,
```
gotmax23, line 93:
```suggestion
    extra_environ: Mapping[str, str] | None = None,
```

src/pyproject_hooks/_impl.py
```diff
@@ -63,7 +87,11 @@ def default_subprocess_runner(cmd, cwd=None, extra_environ=None):
     check_call(cmd, cwd=cwd, env=env)
 
 
-def quiet_subprocess_runner(cmd, cwd=None, extra_environ=None):
+def quiet_subprocess_runner(
+    cmd: Sequence[str],
+    cwd: Optional[str] = None,
```
gotmax23, line 92: See above.

src/pyproject_hooks/_impl.py
```diff
@@ -6,31 +6,51 @@
...
+    from typing import Protocol
 
-def write_json(obj, path, **kwargs):
+    class SubprocessRunner(Protocol):
+        """A protocol for the subprocess runner."""
+
+        def __call__(
+            self,
+            cmd: Sequence[str],
+            cwd: Optional[str] = None,
```
gotmax23, line 22:
```suggestion
            cmd: Sequence[StrPath],
            cwd: Optional[StrPath] = None,
```

This and the rest of the code that accepts file paths should accept PathLike objects. There's a couple ways to annotate this, but my preference is to add `from _typeshed import StrPath` under an `if TYPE_CHECKING` block and use that.

The underlying subprocess code technically supports bytes as well. You can use `StrOrBytes` path for the subprocess_runner parts if you prefer.

src/pyproject_hooks/_impl.py
```diff
@@ -260,10 +317,16 @@ def prepare_metadata_for_build_editable(
         )
 
     def build_editable(
-        self, wheel_directory, config_settings=None, metadata_directory=None
-    ):
+        self,
+        wheel_directory: str,
+        config_settings: Optional[Mapping[str, Any]] = None,
```
gotmax23, line 322:
```suggestion
        config_settings: Mapping[str, Any] | None = None,
```

The `_directory` params should use StrPath.

src/pyproject_hooks/_impl.py
```diff
@@ -6,31 +6,51 @@
 from os.path import abspath
 from os.path import join as pjoin
 from subprocess import STDOUT, check_call, check_output
+from typing import TYPE_CHECKING, Any, Iterator, Mapping, Optional, Sequence
```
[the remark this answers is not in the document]
pradyunsg: I'm not too keen on adopting `from __future__ import annotations` here. We can take that up in a follow up, if needed.

src/pyproject_hooks/_impl.py
gotmax23: [`_typing.StrPath`](https://github.com/python/typeshed/blob/3c872ca8fd875f2dc5fe5f5d771e35c58390cd0e/stdlib/_typeshed/__init__.pyi#L143) is considered [stable](https://github.com/python/typeshed/tree/main/stdlib/_typeshed#api-stability). It is equivalent to `Union[str, "os.PathLike[str]"]` if you prefer to define it manually.

pradyunsg: Let's do this in a follow up change.

pradyunsg:
On the basis of "this is better than status quo", I'm gonna go ahead and merge this.

There's a couple of potential follow ups to improve this further:

- Using the newer typing syntax and more granular "semantic" types, that @gotmax23 has suggested.
- Making the protocol available on Python 3.8+ at runtime (with an object on 3.7), as @henryiii suggested.

pradyunsg:
(I'll fix the documentation build separately)

PR #172, opened 2023-07-03
chore: ruff moved to astral-sh

henryiii:
See https://github.com/scientific-python/cookie/pull/200.

(Also bumped a few versions since I'm touching the file).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,11 +1,11 @@
 repos:
-  - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: v0.0.241
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.0.255
     hooks:
       - id: ruff
 
   - repo: https://github.com/psf/black
-    rev: 23.1.0
+    rev: 23.3.0
     hooks:
       - id: black
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.0.255
+    rev: v0.4.2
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -1,11 +1,11 @@
 repos:
-  - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: v0.0.241
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.0.255
```
takluyver, line 3:
```suggestion
    rev: v0.4.2
```

.pre-commit-config.yaml
```diff
@@ -1,11 +1,11 @@
...
     hooks:
       - id: ruff
 
   - repo: https://github.com/psf/black
-    rev: 23.1.0
+    rev: 23.3.0
```
takluyver, line 8:
```suggestion
    rev: 24.4.2
```

takluyver (commented):
Sorry, I've overlooked this for some time. Trying with updated versions now...

.pre-commit-config.yaml
```diff
@@ -1,11 +1,11 @@
...
-  - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: v0.0.241
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.4.2
     hooks:
       - id: ruff
 
   - repo: https://github.com/psf/black
-    rev: 23.1.0
+    rev: 24.4.2
```
takluyver, line 8:
```suggestion
    rev: 23.1.0
```

Go back to the version that accepts the code as is - we can update the style in another PR.

takluyver:
I had to rebase this to pick up a separate fix so the docs build passes, but now it looks OK. :+1:

PR #182, opened 2023-11-08
0.13.1 release for pep517, with deprecation notice

### LINKED ISSUE #178: Add renamed/deprecated notice to pypi.org/project/pep517
Could you please add a new release of pep517 that links to pyproject-hooks (and build) quite visibly in the long description? Currently, when I go to https://pypi.org/project/pep517/, nothing warns me I should not use that package.

Thanks.

takluyver:
Closes #178

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.13.0
+current_version = 0.13.1
 commit = True
 tag = True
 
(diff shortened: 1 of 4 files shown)


takluyver:
I'm going to merge this, because the goal is to make a release. The remaining failure is on RTD, which doesn't matter, because we don't need to change the docs at all (https://pep517.readthedocs.io/en/latest/  already shows the warning).

PR #184, opened 2024-02-27
Export `SubprocessRunner`

layday:
mypy in strict mode will complain that `SubprocessRunner` is not defined if it is not explicitly exported.

### DIFF
--- a/src/pyproject_hooks/__init__.py
+++ b/src/pyproject_hooks/__init__.py
@@ -12,10 +12,6 @@ from ._impl import (
     quiet_subprocess_runner,
 )
 
-if TYPE_CHECKING:
-    from ._impl import SubprocessRunner
-
-
 __version__ = "1.0.0"
 __all__ = [
     "BackendUnavailable",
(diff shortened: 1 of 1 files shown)


src/pyproject_hooks/__init__.py
```diff
@@ -20,7 +24,5 @@
     "default_subprocess_runner",
     "quiet_subprocess_runner",
     "BuildBackendHookCaller",
+    "SubprocessRunner",
```
takluyver, line 27: I don't think this is compatible with the `if TYPE_CHECKING`  condition - it looks like doing `from pyproject_hooks import *` would fail with an AttributeError when it tries to get `SubprocessRunner`.

Does mypy accept anything like:

```python
if TYPE_CHECKING:
    __all__.append("SubprocessRunner")
```

layday: Ah, good catch. `__all__ += [...]` appears to work.

layday: Honestly forgot `__all__` has an impact on the runtime, lol.

takluyver: :laughing: Thanks!

PR #190, opened 2024-04-29
CI: Update Github checkout & setup-python actions

takluyver:
The CI is issuing warnings about the version of node that these actions use. Updating to the latest versions should fix them.


takluyver:
That seems to have worked; I'll merge this straight away on the grounds that it's boring and not related to the actual code.

PR #196, opened 2024-04-30
Ensure tests, docs, license etc. are included in sdist

### LINKED ISSUE #194: 1.1.0 sdist is missing tests and `LICENSE` file
The 1.1.0 sdist is missing both tests and the `LICENSE` file.  This is regression from the previous release 1.0.0.  Please add both tests and the missing `LICENSE` file to sdist.  Thank you.

takluyver:
They were previously included automatically by `flit build` because they're checked into git, but we're moving away from that to a more explicit listing.

Closes #194

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,6 +8,7 @@ authors = [
     {name = "Thomas Kluyver", email = "thomas@kluyver.me.uk"},
 ]
 readme = "README.rst"
+license = {file = "LICENSE"}
 requires-python = ">=3.7"
 dependencies = []
 classifiers = [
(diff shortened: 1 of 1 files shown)


ofek:
Can this be merged?

jaraco (approved):
LGTM.

Just a reminder that foot guns like these should probably be avoided. IMNSHO, any build system should provide low-risk defaults. Both setuptools with setuptools-scm and [coherent.build](/coherent-oss/coherent.build) provide the behavior (code in repo translates to code in sdist) out of the box without extra configuration.

ofek:
Agreed, Hatchling has such defaults:

- https://hatch.pypa.io/latest/plugins/builder/sdist/#default-file-selection
- https://hatch.pypa.io/latest/plugins/builder/wheel/#default-file-selection

PR #197, opened 2024-04-30
Don't create tag automatically on bumping version number

takluyver:
I need to make the version number change in a PR, but I'd like the tag to point to the merge commit when I merge the PR, so I'll make it manually.


webknjaz:
I like doing this automatically post PyPI release: https://github.com/ansible/awx-plugins/blob/e613d5a/.github/workflows/ci-cd.yml#L828-L923. Since it's the point of no return that the tag wouldn't be deleted if upload happened.

PR #198, opened 2024-04-30
chore: make pre-commit faster

henryiii:
This is recommended as it gives the compiled wheel of black instead of building from source and skipping the mypyc optimizations.

There's also ruff-format, which is already available in the ruff pre-commit hook and could be swapped out if you want.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ repos:
     hooks:
       - id: ruff
 
-  - repo: https://github.com/psf/black
+  - repo: https://github.com/psf/black-pre-commit-mirror
     rev: 23.3.0
     hooks:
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks, I'm happy with this. :+1: 

Do you have opinions on https://pre-commit.ci/ ? It claims to run these checks much faster than doing them as part of normal CI.

henryiii:
I’m a fan, and use it in most repos. It’s fast (though GHA is fast). It’s really nice for the autofix and auto update features (can be opted out).

You need someone with access to all pypa repos to enable it, I think. Unless GitHub fixed the “feature” that disabled all repos that a user didn’t have access to when you modify access. That happened to me about a year ago. 😳

PR #199, opened 2024-05-02
Improve interop with `importlib.metadata`: variation without `importlib_metadata`

### LINKED ISSUE #192: Release 1.1: Regression breaks setuptools builds

abravalheri:
This is basically the same as https://github.com/pypa/pyproject-hooks/pull/195 but without using `importlib_metadata`.

Feel free to close if #195 is preferred.

[This](https://github.com/pypa/pyproject-hooks/pull/195#discussion_r1587336156)  might be worthy considering:

> In general, I find that the following works relatively fine:
>
> * using importlib_metadata.DistributionFinder, MetadataPathFinder to implement find_distrubitions, but then using importlib.metadata high-level APIs
>
> And the following does not work:
> 
> * using importlib.metadata.DistributionFinder, MetadataPathFinder to implement find_distrubitions, but then using importlib_metadata high-level APIs

i.e. while this approach decreases the "spooky action at distance" of a transitive dependency installing `importlib_metadata`, it may also be more error prone if some dependency is indeed using `importlib_metadata`.

Closes #192

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -3,4 +3,3 @@ flake8
 testpath
 setuptools>=30
 tomli >=1.1.0 ; python_version<'3.11'
-importlib-metadata; python_version<'3.8'
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -3,3 +3,4 @@ flake8
 testpath
 setuptools>=30
 tomli >=1.1.0 ; python_version<'3.11'
+importlib-metadata; python_version<'3.8'
--- a/src/pyproject_hooks/_in_process/_in_process.py
+++ b/src/pyproject_hooks/_in_process/_in_process.py
@@ -106,6 +106,13 @@ class _BackendPathFinder:
 
         return spec
 
+    def find_distributions(self, context=None):
+        # Delayed import: Python 3.7 does not contain importlib.metadata
+        from importlib.metadata import DistributionFinder, MetadataPathFinder
+
+        context = DistributionFinder.Context(path=self.backend_path)
+        return MetadataPathFinder.find_distributions(context=context)
+
 
 def _supported_features():
     """Return the list of options features supported by the backend.
--- a/src/pyproject_hooks/_in_process/_in_process.py
+++ b/src/pyproject_hooks/_in_process/_in_process.py
@@ -106,17 +106,14 @@ class _BackendPathFinder:
 
         return spec
 
-    def find_distributions(self, context=None):
-        # Delayed import: Python 3.7 does not contain importlib.metadata
-        # If this method is being called it must be because
-        # `importlib.metadata`/`importlib_metadata` is available.
-        try:
-            from importlib_metadata import DistributionFinder, MetadataPathFinder
-        except ImportError:
+    if sys.version_info >= (3, 8):
+
+        def find_distributions(self, context=None):
+            # Delayed import: Python 3.7 does not contain importlib.metadata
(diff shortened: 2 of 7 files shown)


takluyver:
Thanks! Does this somehow avoid the exception you found in setuptools tests in [this comment](https://github.com/pypa/pyproject-hooks/pull/195#issuecomment-2088695300)?

abravalheri:
For setuptools, it should be fine https://github.com/abravalheri/setuptools/pull/14 (tests should OK, docs failing but that is unrelated).

(I went ahead and already added fixed the `importlib.metadata` compatibility in setuptools: https://github.com/pypa/setuptools/pull/4339)

takluyver:
Aha, so this is functionally similar to the original version of #195 (before involving `importlib_metadata`), but setuptools has changed so that it works?

I prefer this approach. I'll give it a day or so in case anyone else wants to weigh in, but if not then I'd like to merge this and cut another release.

ofek:
What are the next steps for this to be merged?

abravalheri:
Hi @jaraco, @takluyver, is this approach (or the one in https://github.com/pypa/pyproject-hooks/pull/195 which includes `importlib_metadata`), a direction that the project is interested in pursuing or is the project interested in pursuing a different approach?

jaraco:
I unfortunately have paged out the rich context, but I agree that it sounds as if this narrow and surgical fix may be a very solid improvement. Let's go with it for now. Thanks for the diligent investigation and proposal.

jaraco (approved):

nanonyme:
Would it be possible to make a release with this soon? The bug this aims to fix is basically update blocker for pyproject-hooks for any distro-like entities.

icp1994:
And for any distro which has upgraded pyproject-hooks unaware of this cannot update affected packages - [pex](https://github.com/pex-tool/pex/blob/v2.4.0/tox.ini#L207) for example.

jaraco:
@pradyunsg or @takluyver ^

takluyver:
Sorry for the delay.

takluyver:
v1.2 is tagged and on PyPI now - thanks for your patience and for the periodic reminders.

henryiii:
1.2 has broken build's test suite when using `backend-path` (and check-sdist's setuptools test, though I just dropped the setuptools test there). See https://github.com/pypa/build/pull/820, https://github.com/python/importlib_metadata/issues/508, https://github.com/henryiii/check-sdist/pull/55. I've finally narrowed it down to this package's 1.2 release.

takluyver:
:face_exhaling: Sorry.

I have no idea what's the right thing to do here now. #165 was meant to improve how an in-tree backend was loaded, but broke setuptools in some scenarios (#192). This PR was meant to fix that, but it seems that it's broken something in another scenario. :disappointed: Maybe we should have used `importlib_metadata` in some scenarios? But IIUC setuptools also has its own bundled copy, so that doesn't necessarily solve it.

I'm sorely tempted to declare that pyproject-hooks now requires Python 3.10 or above, and you should just pin whichever version works for you on older Python versions. I'm normally pretty keen on backwards compatibility, so I'd welcome a better solution, but I don't have the time or energy to wade through importlib stuff to figure this out. And I don't know how we can trust that the next solution won't just lead to another of these bugs popping up somewhere.

pradyunsg:
Let's have the discussion on that in #206.

PR #202, opened 2024-06-26
Speed up execution of hooks

ofek:
Don't know if this will be accepted but based on the results of `hyperfine` this reduces overall time by ~ 10 ms


ofek:
Not that much of an improvement to justify potential unexpected behavior.

PR #204, opened 2024-09-29
Relax test for build requirements from setuptools

### LINKED ISSUE #203: test_setup fails with setuptools >= 70.1.0
pytest -k "test_setup"

gives

```

    def test_setup_py():
        hooks = get_hooks("setup-py")
        with modified_env({"PYTHONPATH": BUILDSYS_PKGS}):
            res = hooks.get_requires_for_build_wheel({})
        # Some versions of setuptools list setuptools itself here
        res = [x for x in res if x != "setuptools"]
>       assert res == ["wheel"]
E       AssertionError: assert [] == ['wheel']
E         
E         Right contains one more item: 'wheel'
E         Use -v to get more diff

tests/test_call_hooks.py:213: AssertionError
```

it works when downgrading to setuptools 70.0.0

takluyver:
Closes #203.

### DIFF
--- a/tests/test_call_hooks.py
+++ b/tests/test_call_hooks.py
@@ -207,10 +207,9 @@ def test_path_pollution():
 def test_setup_py():
     hooks = get_hooks("setup-py")
     with modified_env({"PYTHONPATH": BUILDSYS_PKGS}):
-        res = hooks.get_requires_for_build_wheel({})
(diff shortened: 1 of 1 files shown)


takluyver:
I'm going to merge this without review, since it only touches tests and fixes the CI.

Issue #168, opened 2023-04-07
Drop support for Python 3.7


pradyunsg:
Python 3.7 goes EOL in 2 months and 3 weeks (https://endoflife.date/python). Filing an issue to track dropping support.

pradyunsg:
For posterity, https://github.com/pypa/pyproject-hooks/commit/69ed8b1dac92e33d6a2d18c1b3a3cfad57020d90 is a part commit working toward this. We should drop the string literal for "SubprocessRunner" and directly document the protocol class as well.

henryiii:
I'd also like the types to be available at runtime for pypa/build, Python 3.7 support is preventing this.

notatallshaw:
This was completed in https://github.com/pypa/pyproject-hooks/pull/214

PR #209, opened 2024-12-26
Correct editable metadata hook return type to str

ichard26:
`prepare_metadata_for_build_wheel` was initially typed to return `Optional[str]` just like the editable metadata hook, but it [got changed to `str` during review](https://github.com/pypa/pyproject-hooks/pull/167#discussion_r1160636565). However, the editable hook's return type was never corrected. This seems to be an oversight as I can't find anything in the PEP 517 or 660 specifications allowing these hooks to return None.

### DIFF
--- a/src/pyproject_hooks/_impl.py
+++ b/src/pyproject_hooks/_impl.py
@@ -285,7 +285,7 @@ class BuildBackendHookCaller:
         metadata_directory: str,
         config_settings: Optional[Mapping[str, Any]] = None,
         _allow_fallback: bool = True,
-    ) -> Optional[str]:
+    ) -> str:
         """Prepare a ``*.dist-info`` folder with metadata for this project.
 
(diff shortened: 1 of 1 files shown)


matthewhughes934:
You could probably drop Python 3.7 support to make the checks green again. It looks like it was previously dropped with 69ed8b1dac92e33d6a2d18c1b3a3cfad57020d90 but that was reverted 9cf1649a071f4268578b0e7417eb017637c4a486 it seems it was kept because `pip` also supported it at the time (context: https://github.com/pypa/pyproject-hooks/pull/167#issuecomment-1500884422) but `pip` has since dropped support for it https://github.com/pypa/pip/commit/1dbd717bb037b62273b7ced3737c4da787515c7c, see also https://github.com/pypa/pyproject-hooks/issues/168

matthewhughes934 (approved):

PR #210, opened 2025-03-31
chore: use subprocess.run instead of old check_*

henryiii:
I noticed this was using the old `subprocess.check_*` methods. In modern Python, `subprocess.run` is [recommended](https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module).

### DIFF
--- a/src/pyproject_hooks/_impl.py
+++ b/src/pyproject_hooks/_impl.py
@@ -2,10 +2,10 @@ import json
 import os
 import sys
 import tempfile
+import subprocess
 from contextlib import contextmanager
 from os.path import abspath
 from os.path import join as pjoin
-from subprocess import STDOUT, check_call, check_output
 from typing import TYPE_CHECKING, Any, Iterator, Mapping, Optional, Sequence
 
 from ._in_process import _in_proc_script_path
@@ -74,13 +74,13 @@ def default_subprocess_runner(
 ) -> None:
     """The default method of calling the wrapper subprocess.
 
-    This uses :func:`subprocess.check_call` under the hood.
+    This uses :func:`subprocess.run` under the hood.
     """
     env = os.environ.copy()
     if extra_environ:
         env.update(extra_environ)
 
-    check_call(cmd, cwd=cwd, env=env)
+    subprocess.run(cmd, cwd=cwd, env=env, check=True)
 
 
 def quiet_subprocess_runner(
@@ -90,13 +90,21 @@ def quiet_subprocess_runner(
 ) -> None:
     """Call the subprocess while suppressing output.
 
-    This uses :func:`subprocess.check_output` under the hood.
+    This uses :func:`subprocess.run` with `PIPE` under the hood.
     """
     env = os.environ.copy()
     if extra_environ:
         env.update(extra_environ)
 
-    check_output(cmd, cwd=cwd, env=env, stderr=STDOUT)
+    subprocess.run(
+        cmd,
+        cwd=cwd,
+        env=env,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+        check=True,
+        text=True,
+    )
 
 
 def norm_and_check(source_tree: str, requested: str) -> str:


src/pyproject_hooks/_impl.py
```diff
@@ -95,13 +95,21 @@ def quiet_subprocess_runner(
...
 
-    check_output(cmd, cwd=cwd, env=env, stderr=STDOUT)
+    subprocess.run(
+        cmd,
+        cwd=cwd,
+        env=env,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+        check=True,
+        text=True,
```
pradyunsg, line 111: Using `text=True` would be a behaviour change.

pradyunsg: Let's drop this line, to be cautious.

src/pyproject_hooks/_impl.py
```diff
@@ -95,13 +95,21 @@ def quiet_subprocess_runner(
 ) -> None:
     """Call the subprocess while suppressing output.
 
-    This uses :func:`subprocess.check_output` under the hood.
+    This uses :func:`subprocess.run` with `PIPE` under the hood.
```
pradyunsg, line 98:
```suggestion
    This uses :func:`subprocess.run` with `stdout=PIPE, stderr=STDOUT` under the hood.
```

src/pyproject_hooks/_impl.py
henryiii:
```suggestion
```

src/pyproject_hooks/_impl.py
```diff
@@ -95,13 +95,20 @@ def quiet_subprocess_runner(
 ) -> None:
     """Call the subprocess while suppressing output.
 
-    This uses :func:`subprocess.check_output` under the hood.
+    This uses :func:`subprocess.run` with `stdout=PIPE, stderr=STDOUT` under the hood.
```
takluyver:
```suggestion
    This uses :func:`subprocess.run` with `stdout=DEVNULL, stderr=DEVNULL` under the hood.
```

henryiii: Does that mean it won't show up in the error? I think you can see it in the error exception because this has `check=True`.

henryiii:
This removes the ability to check and see why a command failed! My version:

```pycon
>>> try:
...     subprocess.run(["ruff", "invalid"], stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
... except subprocess.CalledProcessError as err:
...     print(err.stdout, err.stderr)
...
b'' b"error: unrecognized subcommand 'invalid'\n\nUsage: ruff [OPTIONS] <COMMAND>\n\nFor more information, try '--help'.\n"
```

Current version:

```pycon
>>> try:
...     subprocess.run(["ruff", "invalid"], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, check=True)
... except subprocess.CalledProcessError as err:
...     print(err.stdout, err.stderr)
...
None None
```

If a command fails, users will want to know why it failed!

This was available before (though with stdout and stderr merged, I didn't do that in my example here):

```pycon
>>> try:
...     subprocess.check_output(["ruff", "invalid"], stderr=subprocess.STDOUT)
... except subprocess.CalledProcessError as err:
...     print(err.stdout, err.stderr)
...
b"error: unrecognized subcommand 'invalid'\n\nUsage: ruff [OPTIONS] <COMMAND>\n\nFor more information, try '--help'.\n" None
```

We catch and print this all the time in build: https://github.com/search?q=repo%3Apypa%2Fbuild%20CalledProcessError&type=code

takluyver:
You're right, my mistake. I figured that if the return value wasn't used,
there was no way to get the output, but I forgot about the exception having
it.

Please knock my commit off the branch. If you want to add a comment & a
test, it might stop me making the same mistake in a couple of years time. 😉

takluyver:
Thanks!

Issue #157, opened 2022-11-18
Enable forwarding warnings from subprocess to the caller


pradyunsg:
This would be a reasonable mechanism to surface deprecation/build-related warnings from underlying build-backends to the build-frontends, which can then surface them to users appropriately.

This would likely involve some way to capture and communicate the log messages back to the calling process; ideally with the ability to show the warnings without waiting for the end of the build-backend subprocess.

pradyunsg:
To be clear, I'm thinking of `warnings.warn`-style warnings.

pradyunsg:
FYI: I've started exploring implementing this and figuring out what this should look like.

pradyunsg:
Hmm... I came here to close this now that #213 is merged but it looks like this might actually need a PEP or something to make things clearer?

@notatallshaw Any concerns if I revert #213 and we resurrect that discussion? I might have missed context from a more recent discussion, if there is one.

notatallshaw:
> [@notatallshaw](https://github.com/notatallshaw) Any concerns if I revert [#213](https://github.com/pypa/pyproject-hooks/pull/213) and we resurrect that discussion? I might have missed context from a more recent discussion, if there is one.

The comment here suggested this would just need a spec clarification not a PEP: https://github.com/pypa/packaging-problems/issues/558#issuecomment-981541735

There was some recent discussion in this thread: https://discuss.python.org/t/how-can-build-backends-avoid-breaking-users-when-they-make-backwards-incompatible-changes/85847

My motivation here is to help build backends communicate with users to reduce the chance of unexpected failures, and I don't really have an opinion on the spec, I haven't read it in detail. But if you're not comfortable with it I am motivated to do whatever steps are required until you are.

pfmoore:
> The comment here suggested this would just need a spec clarification not a PEP

So that means it needs (1) a dedicated thread on Discourse proposing the change and establishing community consensus that it's ok as a spec change, and (2) the actual spec change having been made.

I agree with @pradyunsg - let's revert until that has happened.

notatallshaw:
> So that means it needs (1) a dedicated thread on Discourse proposing the change and establishing community consensus that it's ok as a spec change, and (2) the actual spec change having been made.

I was being optimistic that this had already been done since your discussion.

I started take a look at doing this in the spec page, but as best as I can tell PEP 517 has never been added to: https://packaging.python.org/en/latest/specifications/

[...]

PR #212, opened 2025-04-10
Fix find_distributions with non-empty context

### LINKED ISSUE #211: Issue with new find_distributions and incorrect metadata
After updating to pyproject_hooks 1.2.0, I am getting the following error when building [setuptools_scm](https://github.com/pypa/setuptools-scm):
```
> /usr/bin/python3.9 -m build -n
* Getting build dependencies for sdist...
/scratch/userland-gate/components/python/setuptools_scm/setuptools_scm-8.1.0/src/setuptools_scm/_integration/setuptools.py:31: RuntimeWarning:
ERROR: setuptools==8.1.0 is used in combination with setuptools_scm>=8.x
```
even though my setuptools version is 73.0.1. The build process erroneously takes the setuptools_scm version and considers it the setuptools one.

(while this says ERROR, the build continues and finishes successfully; the only unexpected result is that the WHEEL file lists `Generator` as `setuptools (8.1.0)`, which is not the case).

I also noticed similar issue with [Python incremental](https://github.com/twisted/incremental).

I know very little about how the whole build machinery works, but I looked a little into this and found that the `def find_distributions` addition in 1.2.0 is the reason for this. It seems that importlib_metadata (which is used by setuptools to find its own `__version__`) uses `_BackendPathFinder` and the new `find_distributions`, but the passed `context` (which is correctly set to look for "setuptools") is ignored and replaced with new context, which is no longer correct for that case.

Updating the function with a check:
```python
if context is None:
    context = DistributionFinder.Context(path=self.backend_path)
```
fixes the issue, but I have no idea whether that is a good idea - it was just a test.

This happens in 3.9.21 (with importlib_metadata 7.1.0), 3.11.11 and 3.13.1.

Also, both setuptools_scm and incremental use custom build backend and "backend-path" (with more than just ".").

kulikjak:
`find_distributions` now honors parameters from the original `context`.

Closes #211.

### DIFF
--- a/src/pyproject_hooks/_in_process/_in_process.py
+++ b/src/pyproject_hooks/_in_process/_in_process.py
@@ -112,7 +112,11 @@ class _BackendPathFinder:
             # Delayed import: Python 3.7 does not contain importlib.metadata
             from importlib.metadata import DistributionFinder, MetadataPathFinder
 
-            context = DistributionFinder.Context(path=self.backend_path)
+            if context is None:
+                context = DistributionFinder.Context(path=self.backend_path)
+            else:
+                args = dict(vars(context), path=self.backend_path)
+                context = DistributionFinder.Context(**args)
             return MetadataPathFinder.find_distributions(context=context)
 
 


src/pyproject_hooks/_in_process/_in_process.py
```diff
@@ -112,7 +112,11 @@ def find_distributions(self, context=None):
             # Delayed import: Python 3.7 does not contain importlib.metadata
             from importlib.metadata import DistributionFinder, MetadataPathFinder
 
-            context = DistributionFinder.Context(path=self.backend_path)
+            if context is None:
+                context = DistributionFinder.Context(path=self.backend_path)
+            else:
+                args = dict(vars(context), path=self.backend_path)
+                context = DistributionFinder.Context(**args)
```
abravalheri, line 119: I think this is good. Thank you.
I had a deeper look into the implementation of `MetadataPathFinder`, and it seems that it only ever looks into [`context.name` and `context.path`](https://github.com/python/cpython/blob/v3.9.21/Lib/importlib/metadata.py#L505), so, *if we want*, this could be further simplified to:

```suggestion
            name = context.name if context else None
            context = DistributionFinder.Context(name=name, path=self.backend_path)
```

abravalheri: I wonder if there is any type of regression test that could be added.

pradyunsg:
Would it be possible to add a test for this, to help ensure we don't regress on this behaviour in the future?

kulikjak:
Sure, I added a new test (and found a problem in existing one).

I didn't have much time to look into it up until today, so it took longer.

kulikjak:
Small update: I now tested the patched version with build of ~120 different Python packages and everything seems to work as expected.

takluyver:
I'll merge this, but I'd like to sort out the questions about warnings (#157) before making a release, to avoid making anything worse.

takluyver:
Thanks @kulikjak !

PR #218, opened 2025-06-03
feat: use `sphinx-issues` for linking

shenxianpeng:
It could be helpful for users if the [Changlog](https://pyproject-hooks.readthedocs.io/en/latest/changelog.html) included links to the related issues or pull requests. This would make it easier to trace the context behind each change.

---

Here used sphinx_issues, and it also aligns with other pypa projects 

https://github.com/pypa/build/blob/6653cd288a4445bdc1ad55968ad584e0eeebc5aa/docs/conf.py#L38

https://github.com/pypa/pip/blob/c6f1c7905eab73f6bbebb0a35db7beeb901c4c36/docs/html/conf.py#L35

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -16,7 +16,6 @@ author = "Thomas Kluyver"
 extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.intersphinx",
-    "sphinx.ext.extlinks",
     "sphinx_issues",
 ]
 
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -46,6 +47,4 @@
 # -- Options for extlinks ----------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/extensions/extlinks.html#configuration
 
-extlinks = {
-    "pypi": ("https://pypi.org/project/%s", "%s"),
```
takluyver, line 50: IIUC, the `:pypi:` role is also (randomly) provided by sphinx-issues? If we're no longer using extlinks at all, shall we remove it from the extensions list to make that clear?

takluyver:
Thanks!

PR #222, opened 2026-02-03
Use os.path.commonpath() instead of commonprefix()

sethmlarson:
Now that `pyproject-hooks` supports only Python 3.6+ `commonprefix()` is no longer required.


jaraco:
Would be nice to see a release of this. Because it's unreleased and it's breaking Setuptools' test suite, we'll have to suppress the warning until a release is cut (and then remove the suppression later).

(to be fair, Setuptools' own warnings-as-errors setting is also implicated)

jaraco:
Closes #221 and #223

notatallshaw:
FWIW, we found this switch created a breaking regression on pip: https://github.com/python/cpython/issues/149444 / https://github.com/pypa/pip/pull/14001

And the solution in the pip PR has a regression, that no one has reported yet, where `/` incorrectly can't be selected.

PR #225, opened 2026-03-27
Pin GitHub Actions and Python packages

sethmlarson:
Hello! This PR pins GitHub Actions and Python packages in use during the release process. As a follow-up, if you'd like to use Dependabot to keep these actions up-to-date I can create a policy that hopefully won't pester you with update PRs.


takluyver:
Thanks, I'm happy to have them updated by dependabot as well.

PR #227, opened 2026-06-04
Document subprocess_runner threading limitation

### LINKED ISSUE #226: BuildBackendHookCaller.subprocess_runner() context manager is not thread-safe
## Description

`BuildBackendHookCaller.subprocess_runner()` context manager mutates `self._subprocess_runner` on the instance without any synchronization:

```python
@contextmanager
def subprocess_runner(self, runner):
    prev = self._subprocess_runner
    self._subprocess_runner = runner
    try:
        yield
    finally:
        self._subprocess_runner = prev
```

If two threads share the same `BuildBackendHookCaller` instance and one of them uses `subprocess_runner()` to temporarily swap the runner, the other thread sees the swapped runner and may invoke hooks with the wrong runner. The restore in `finally` also races with concurrent reads of `self._subprocess_runner` in `_call_hook`.

## Impact

This prevents consumers (e.g. `pypa/build`) from safely sharing a `BuildBackendHookCaller` (or a wrapping `ProjectBuilder`) across threads when `subprocess_runner()` is used.

Discovered during a thread-safety audit of `pypa/build` for https://github.com/pypa/build/issues/377.

Himanshuagrawal4:
## Summary
- Document that BuildBackendHookCaller.subprocess_runner() temporarily mutates the hook caller instance.
- Note that the context manager is not thread-safe and suggest separate caller instances or constructor-level runners for concurrent use.

## Tests
- git diff --check
- python -m pytest tests\\test_call_hooks.py -q (19 passed, isolated temp venv)

Addresses #226.

### DIFF
--- a/src/pyproject_hooks/_impl.py
+++ b/src/pyproject_hooks/_impl.py
@@ -175,6 +175,13 @@ class BuildBackendHookCaller:
 
         :param runner: The new subprocess runner to use within the context.
 
+        .. warning::
+
+            This context manager temporarily mutates the hook caller instance and
+            is not thread-safe. Callers that need to run hooks concurrently
+            should create separate :class:`BuildBackendHookCaller` instances, or
+            provide the subprocess runner when constructing the caller.
+
         .. code-block:: python
 
             hook_caller = BuildBackendHookCaller(...)


yangfan-yf-yf (approved):
﻿Reviewed `608b1c6` independently.

The warning accurately describes the current context-manager behavior and follows the maintainer decision in #226: document the mutable, non-thread-safe scope rather than introduce implicit thread-local behavior. The suggested alternatives match the existing constructor API.

Validation on Windows with Python 3.12:

- `python -m pytest -q` ? 40 passed
- `git diff --check` ? clean

The Read the Docs build is also passing.

takluyver:
Thanks!

PR #229, opened 2026-06-09
Add security policy

### LINKED ISSUE #228: Add a security policy
Please add a security policy (and possibly enable GHSAs as a way to accept reports), as currently it is unclear where private vulnerability reports should go. As a result, they end up in the PSRT's mailbox. If you need help reviewing any such policies, feel free to ping me :-)

Himanshuagrawal4:
## Summary
- add a SECURITY.md for pyproject-hooks
- point security reporters to PyPA's security reporting guidelines
- note that security issues should not be reported in the public issue tracker

Closes #228.

## Validation
- git diff --check

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,6 @@
+Security Policy
+===============
+
+To report a security issue, please follow the special [security reporting
+guidelines](https://packaging.python.org/en/latest/security/). Do not report
+security issues in the public issue tracker.


SECURITY.md
```diff
@@ -0,0 +1,6 @@
+Security Policy
+===============
+
+To report a security issue, please follow the special [security reporting
+guidelines](https://packaging.python.org/en/latest/security/). Do not report
```
pfmoore, line 5: This link doesn't exist.

notatallshaw: Yes, this PR is nonsense, PyPA has no organizational level security reporting process, it is up to individual projects to determine their security policy. 

I would strongly suggest that only maintainers should open a PR to fix https://github.com/pypa/pyproject-hooks/issues/228

Himanshuagrawal4:
Thanks for the review. You're right: the link I used is not valid for this project, and setting a PyPA-level reporting policy is a maintainer-owned decision. I'm closing this PR to avoid adding misleading security guidance. Sorry for the noise.

PR #230, opened 2026-07-03
Handle symlinked _in_process path cleanup

### LINKED ISSUE #207: test_path_pollution fails in venv-based environment
`test_path_pollution` (added with `1.2.0`) fails with:
```
_____________________________ test_path_pollution ______________________________

    def test_path_pollution():
        hooks = get_hooks("path-pollution")
        with TemporaryDirectory() as outdir:
            with modified_env(
                {
                    "PYTHONPATH": BUILDSYS_PKGS,
                    "TEST_POLLUTION_OUTDIR": outdir,
                }
            ):
                hooks.get_requires_for_build_wheel({})
            with open(pjoin(outdir, "out.json")) as f:
                captured_sys_path = json.load(f)
    
        with in_proc_script_path() as path:
            assert os.path.dirname(path) not in captured_sys_path
>       assert captured_sys_path[0] == BUILDSYS_PKGS
E       AssertionError: assert '/usr/src/RPM...s/_in_process' == '/usr/src/RPM...buildsys_pkgs'
E         
E         - /usr/src/RPM/BUILD/python3-module-pyproject_hooks-1.2.0/tests/samples/buildsys_pkgs
E         + /usr/src/RPM/BUILD/python3-module-pyproject_hooks-1.2.0/.run_venv/lib/python3/site-packages/pyproject_hooks/_in_process
```

According to the description of f71a8f5399859a9a989c51cd13a456f21f98d947 this test checks if `_in_process` path is not on a head of `sys.path`.

In virtual environment created with stdlib's `venv` `lib64` directory is a link to `lib`, while `virtualenv` (the default backend for `nox`) does create both `lib64` and `lib` directories.

In the code for removal of `_in_process` https://github.com/pypa/pyproject-hooks/blob/main/src/pyproject_hooks/_in_process/_in_process.py#L363-L365
`__file__` contains not resolved path. But during a hook call (`python abspath(str(script))`) `getcwd()` is called which does resolve path and that path is inserted into beginning of `sys.path`.

In virtual environments containing `lib64` symlink the `_in_process` path's parent directory is not cleaned up from `sys.path`.

itscloud0:
## Summary
- fix `_in_process.py` path cleanup to remove `sys.path` entries by normalized real path instead of exact string match
- add a regression that exercises hook calls through a symlinked `_in_process.py` path

## Problem
Issue #207 reports that in venv layouts with a symlinked library path (for example `lib64 -> lib`), the `_in_process` helper can leave its own directory at the front of `sys.path`. That pollutes backend imports instead of restoring the build-system path to the front.

The root cause is that `__file__` keeps the symlink spelling while Python inserts the resolved real path into `sys.path[0]`, so the current exact-string removal misses the injected directory.

## Verification
- `python -m pytest -q`
- `ruff check .`
- `black --check .`
- `pre-commit run mypy --all-files`

Closes #207.

### DIFF
--- a/src/pyproject_hooks/_in_process/_in_process.py
+++ b/src/pyproject_hooks/_in_process/_in_process.py
@@ -362,9 +362,10 @@ def main():
 
     # Remove the parent directory from sys.path to avoid polluting the backend
     # import namespace with this directory.
-    here = os.path.dirname(__file__)
-    if here in sys.path:
-        sys.path.remove(here)
+    here = os.path.normcase(os.path.realpath(os.path.dirname(__file__)))
+    sys.path[:] = [
+        path for path in sys.path if os.path.normcase(os.path.realpath(path)) != here
+    ]
 
     hook = globals()[hook_name]
 
--- a/tests/test_call_hooks.py
+++ b/tests/test_call_hooks.py
@@ -1,3 +1,4 @@
+from contextlib import contextmanager
 import json
 import os
 import tarfile
@@ -10,6 +11,7 @@ import pytest
 from testpath import assert_isfile, modified_env
 from testpath.tempdir import TemporaryDirectory, TemporaryWorkingDirectory
 
+import pyproject_hooks._impl as impl
 from pyproject_hooks import (
     BackendUnavailable,
     BuildBackendWarning,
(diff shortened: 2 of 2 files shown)


takluyver:
Thanks!

PR #235, opened 2026-09-08
Prepare to release 1.3

takluyver:


takluyver:
Great, 1.3 coming soon to an index near you, then. :slightly_smiling_face:

takluyver:
:face_exhaling: I had to mess around with Github actions to fix the upload, and in haste I created new git tags but didn't bump the version number. So it's on PyPI as 1.3.0, but the Github tag says 1.3.2 (with the exact same code).

I'm sure that's going to cause someone a headache, so I'll sort it out one way or another, but I'm out of time to do it just now.

takluyver:
The tag and the PyPI release are now aligned again at 1.3.3. There are no code changes from 1.3.0.

PR #236, opened 2026-09-15
docs: build_sdist returns a name within sdist_directory

hxperl:
`build_sdist`'s `:returns:` names `wheel_directory`, which is not one of its parameters — the directory it writes to is `sdist_directory`. Copy-paste from `build_wheel` directly above it, introduced in 0ee69f40.

It reaches the rendered page. `sphinx -b text docs`, on unmodified `main`:

```
build_sdist(sdist_directory, config_settings=None)

   Build an sdist from this project.

   Returns:
      The name of the newly created sdist within "wheel_directory".
```

One word changed; that line now reads `"sdist_directory"`.

Docs build clean with `-W`, `pytest` 42 passed, pre-commit passes — macOS arm64, CPython 3.14.


takluyver:
Thanks!

